PR #9, opened 2017-03-28
Add ReplaceMatchingSequence extension method.

Fumblesneeze:
usage example:

```csharp
    [HarmonyPatch(typeof(Verb))]
    [HarmonyPatch("TryCastNextBurstShot")]
    public class BurstShotPatch
    {
        public static float _modifier = 1 / 2;

        public static IEnumerable<CodeInstruction> Transpiler(IEnumerable<CodeInstruction> source)
        {
            var field = AccessTools.Field(typeof(VerbProperties), nameof(VerbProperties.ticksBetweenBurstShots));
            var modifierField = AccessTools.Field(typeof(BurstShotPatch), nameof(BurstShotPatch._modifier));
            return source.ReplaceMatchingSequence(
                new Func<CodeInstruction, bool>[]
                {
                    (o => o.operand == field)
                },
                i => i.Concat(
                    new[]
                    {
                        new CodeInstruction(OpCodes.Conv_R4),
                        new CodeInstruction(OpCodes.Ldsfld, modifierField),
                        new CodeInstruction(OpCodes.Mul),
                        new CodeInstruction(OpCodes.Conv_I)
                    }
                )
            );
        }
    }
```

### DIFF
--- a/Harmony/Tools/Extensions.cs
+++ b/Harmony/Tools/Extensions.cs
@@ -36,13 +36,15 @@ namespace Harmony
 		}
 	}
 
-	public static class CollectionExtensions
+	public static class EnumerableExtensions
 	{
 		public static IEnumerable<T> Do<T>(this IEnumerable<T> sequence, Action<T> action)
 		{
 			if (sequence == null) return null;
-			var enumerator = sequence.GetEnumerator();
-			while (enumerator.MoveNext()) action(enumerator.Current);
+			foreach (var item in sequence)
+			{
+				action(item);
+			}
 			return sequence;
 		}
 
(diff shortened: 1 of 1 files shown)


pardeike:
I would love to pull this but please show me how you think this would work if you want to replace like let’s say three or four consecutive codes with something else. I get the replacement but not the way you would specify the matching in a convenient way.

Fumblesneeze:
The matcher is just an array of predicates. so you can just throw this into the matcher:
```                new Func<CodeInstruction, bool>[]
                {
                    o => o.operand == field,
                    _ => true,
                    _ => true
                },
```
to match the field opcode and the two following codes.

I've done this here: https://github.com/Fumblesneeze/RimBetterSpeed/blob/master/Source/BetterSpeed/WarmupPatch.cs
Although in that case I wanted to match with the field opcode because it's easy to do and then add my own codes two opcodes later.

If you meant consecutive similar Matches, e.g. if the same field gets matched several times in a method, then you can use a local variable in the replace Lambda like so:

```
        public static IEnumerable<CodeInstruction> Transpiler(IEnumerable<CodeInstruction> source)
        {
            var field = AccessTools.Field(typeof(VerbProperties), nameof(VerbProperties.ticksBetweenBurstShots));
            var modifierField = AccessTools.Field(typeof(BurstShotPatch), nameof(BurstShotPatch._modifier));
            var count = 0;
            return source.ReplaceMatchingSequence(
                new Func<CodeInstruction, bool>[]
                {
                    (o => o.operand == field)
                },
                i => {
                    if(count++ == 0)
                        return i.Concat(new[]
                        {
                            new CodeInstruction(OpCodes.Conv_R4),
                            new CodeInstruction(OpCodes.Ldsfld, modifierField),
                            new CodeInstruction(OpCodes.Mul),
                            new CodeInstruction(OpCodes.Conv_I)
                        }
                    else return i;
                }
                )
            );
        }
```
that would only append to the first instance of the field.

pardeike:
Ok, I like the flexibility. But I don't like the API. Have a look at what I did in Zombieland:
```csharp
[HarmonyPatch(typeof(Pawn_InteractionsTracker))]
[HarmonyPatch("TryInteractWith")]
static class Pawn_InteractionsTracker_TryInteractWith_Patch
{
	static IEnumerable<CodeInstruction> Transpiler(ILGenerator generator, MethodBase method, IEnumerable<CodeInstruction> instructions)
	{
		var conditions = new List<CodeInstruction>();
		conditions.AddRange(Tools.NotZombieInstructions(generator, method));
		conditions.AddRange(Tools.NotZombieInstructions(generator, method, "recipient"));
		var transpiler = Tools.GenerateReplacementCallTranspiler(conditions);
		return transpiler(generator, instructions);
	}
}
```

Not the actual functionality but more the way I designed the API. This is a transpiler factory and thus can be configured with several calls. Don't you think this makes things better to read?

Fumblesneeze:
As this is literally working with an enumeration I prefer working with LINQ-esque extensions to IEnumerable to modify the whole collection. That way you can still chain it with regular LINQ methods like .Where() etc. It is intuitive to me, as I'm used to working that way (and I would say that is the case for a lot of C# devs but I have no data for that).

Based on this replace method, you could obviously create more specialized methods as wrappers, for example ``` ReplaceMethodCall(this IEnumerable<..> source, MethodInfo from, MethodInfo to) ```
And if you have different Jobs just chain the methods and if you need to do some weird stuff there's always the possibility to foreach even after using these methods.

The way I see it this transpilergenerator does the same stuff only it's saved as a seperate object instead of saved in the IEnumerable directly.

PR #12, opened 2017-04-21
Make MethodInvoker.Handler operate on stack instead of locals

0x0ade:
MethodInvoker.Handler uses locals as a layer of indirection for the arguments, which is useful for reference types (`ref` and `out` parameters), but adds some overhead when invoking the method.

Recently, I've spent my time implementing `ref` support to the ReflectionHelper in MonoMod (offline .NET assembly patcher), but instead of using locals, I opted to operate on the stack. This means MonoMod's ReflectionHelper is now basically supporting everything Harmony's MethodInvoker does.

It's not as clear and took some time to figure out, but after benchmarking the results in MonoMod, it was visible that it's profitable: 

```
100000000 runs:
dmdNewBox (ReflectionHelper modified, immutable boxes): 36.5712269s
dmdMutable (ReflectionHelper, formerly called raw): 16.3218110s
dmdLocals (MethodInvoker): 46.7835971s

Calculated using ticks:
NewBox / Locals: 78.1710453384526%
Locals / NewBox: 127.924603754543%

NewBox / Mutable: 224.063536209309%
Mutable / NewBox: 44.6301980642602%

Locals / Mutable: 286.632390854177%
Mutable / Locals: 34.887892363454%
```

For backwards compatibility, I've set `directBoxValueAccess` to false by default. With it set to true, if passing a ValueType as `ref` argument, all other instances of the box are affected, too. While disabling this "safety" adds quite a performance boost when compared to the original local-using approach, I don't know if enabling raw box value access is feasible in Harmony.

It was interesting to see that there still was a slightly notable performance boost when cutting out locals and operating with the passed array directly.

I don't want to eat the cake alone, so take a piece :)

### DIFF
--- a/Harmony/Extras/MethodInvoker.cs
+++ b/Harmony/Extras/MethodInvoker.cs
@@ -8,7 +8,7 @@ namespace Harmony
 
 	public delegate object FastInvokeHandler(object target, object[] paramters);
 
-	class MethodInvoker
+	public class MethodInvoker
 	{
 		public static FastInvokeHandler GetHandler(DynamicMethod methodInfo, Module module)
 		{
--- a/Harmony/Extras/MethodInvoker.cs
+++ b/Harmony/Extras/MethodInvoker.cs
@@ -20,43 +20,82 @@ namespace Harmony
 			return Handler(methodInfo, methodInfo.DeclaringType.Module);
 		}
 
-		static FastInvokeHandler Handler(MethodInfo methodInfo, Module module)
+		static FastInvokeHandler Handler(MethodInfo methodInfo, Module module, bool directBoxValueAccess = false)
 		{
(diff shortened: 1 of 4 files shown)


pardeike:
To be honest, when I added MethodInvoker I copied it from an older project and never got around using or testing it. So it's kind of hard for me to determine the changes you made. So my question is: would you mind writing at least some test for this or are you confident that I can pull this in without breaking existing behaviour for those few using this class?

0x0ade:
I've just added two tests for `MethodInvoker`. I had to make `MethodInvoker` public for this.

Those tests also pass with the old `MethodInvoker` implementation for me. The only behavioral change noticeable for callers would be that passing a `struct` / `ValueType` as "this" now unboxes "this", instead of force-passing the boxed reference as "this".

pardeike:
Awesome. Good work.

Issue #2, opened 2017-02-06
Harmony should support MS .NET and not only Mono. (was: Error trying to override int getter)


llaughlin:
Target class:
```
public sealed class VisualFind
{
    private int _waitForElementsTimeout;

    public int WaitOnElementsTimeout
    {
      get
      {
        return this._waitForElementsTimeout;
      }
      set
      {
        this._waitForElementsTimeout = value;
    }
}
```

Patcher:
```
internal class HarmonyPatcher
{
        private static readonly Logger _Log = LogManager.GetCurrentClassLogger();

        public static HarmonyInstance Initialize()
        {
            _Log.Info($"Initializing harmony patches");
            HarmonyInstance harmony = HarmonyInstance.Create("com.test");
            harmony.PatchAll(Assembly.GetExecutingAssembly());
            return harmony;
        }
}
```

Patch:
```
[HarmonyPatch(typeof(VisualFind))]
[HarmonyPatch("get_WaitOnElementsTimeout")]
public class VisualFindPatch
{
        private static readonly Logger _Log = LogManager.GetCurrentClassLogger();

        [HarmonyPrefix]
        static void BeforeGet()
        {

        }
}
```

I get the following error: 
```
2017-02-06 14:18:45.0133 INFO Test.Ui.Patches.HarmonyPatcher.Initialize(Main.cs:19)   Initializing harmony patches
2017-02-06 14:18:45.0653 FATAL Test.Ui.Utility.Test.BaseTest.Initialize(BaseTest.cs:114) System.NotSupportedException: Wrong MethodAttributes or CallingConventions for DynamicMethod. Only public, static, standard supported
   at System.Reflection.Emit.DynamicMethod.CheckConsistency(MethodAttributes attributes, CallingConventions callingConvention)
   at System.Reflection.Emit.DynamicMethod.Init(String name, MethodAttributes attributes, CallingConventions callingConvention, Type returnType, Type[] signature, Type owner, Module m, Boolean skipVisibility, Boolean transparentMethod, StackCrawlMark& stackMark)
   at System.Reflection.Emit.DynamicMethod..ctor(String name, MethodAttributes attributes, CallingConventions callingConvention, Type returnType, Type[] parameterTypes, Type owner, Boolean skipVisibility)
   at Harmony.DynamicTools.CreateDynamicMethod(MethodBase original, String suffix) in C:\repo\Harmony\Harmony\Tools\DynamicTools.cs:line 22
[...]

PR #14, opened 2017-05-01
.NET Framework support

0x0ade:
Relevant: #2 

`DynamicMethod::CreateDynMethod` doesn't exist in the .NET Framework; instead, `DynamicMethod::GetMethodDescriptor` and `RuntimeHelper::_CompileMethod` do.

As `_CompileMethod` seems to float around with different signatures (can take either the `RuntimeMethodHandle` or its `IntPtr` `Value` as only argument), `PrepareDynamicMethod` takes both variants into account.

I still need to get a test case mod running; @llaughlin, it'd be nice if you could help testing this as you seem to already have a test case :)

Notes:
* I'm currently manually reformatting the code to fit your existing code style.  
* All tests are failing locally without a copied `0Harmony.dll`, which is why I set
```xml
<Private>True</Private>
```
in `HarmonyTests.csproj`. Some tests are still failing, f.e. the `AccessCache_Method` test seems to fail due to an internal change in Harmony I guess?

I don't deem this PR to be complete until we get something detoured and running :)

### DIFF
--- a/Harmony/ILCopying/Memory.cs
+++ b/Harmony/ILCopying/Memory.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Reflection;
 using System.Reflection.Emit;
+using System.Runtime.CompilerServices;
 
 namespace Harmony.ILCopying
 {
@@ -25,21 +26,23 @@ namespace Harmony.ILCopying
 
 		private readonly static FieldInfo f_DynamicMethod_m_method =
 			// .NET
-			typeof(DynamicMethod).GetField("m_method", BindingFlags.NonPublic | BindingFlags.Instance);
+			typeof(DynamicMethod).GetField("m_method", BindingFlags.NonPublic | BindingFlags.Instance) ??
+			// Mono
+			typeof(DynamicMethod).GetField("mhandle", BindingFlags.NonPublic | BindingFlags.Instance);
 		public static long GetMethodStart(MethodBase method)
 		{
 			RuntimeMethodHandle handle;
 
-			if (method is DynamicMethod)
-			{
-				if (f_DynamicMethod_m_method != null)
-					handle = (RuntimeMethodHandle) f_DynamicMethod_m_method.GetValue(method);
-				else
-					handle = method.MethodHandle;
-			}
+			if (method is DynamicMethod && f_DynamicMethod_m_method != null)
+				handle = (RuntimeMethodHandle) f_DynamicMethod_m_method.GetValue(method);
 			else
 				handle = method.MethodHandle;
 
+			/* Required to ensure that the method is already JITed and the method start doesn't change.
+			 * This seemingly only affects the .NET Framework.
+			 * - ade
+			 */
+			RuntimeHelpers.PrepareMethod(handle);
 			return handle.GetFunctionPointer().ToInt64();
 		}
 
--- a/Harmony/ILCopying/Memory.cs
+++ b/Harmony/ILCopying/Memory.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Reflection;
+using System.Reflection.Emit;
 
 namespace Harmony.ILCopying
 {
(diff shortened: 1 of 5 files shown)


pardeike:
Interesting work, I'll have a look at it as soon as I have time!

llaughlin:
There's nothing special about my issue (https://github.com/pardeike/Harmony/issues/2) that would require a separate test case. The existing `AccessToolsClass` or `AllAttributesClass` classes should be sufficient, to my knowledge.

0x0ade:
I just found the commented `TestMethod1` test and got it running on .NET Framework... when debugging in Visual Studio.

I've had to make `Memory.GetMethodStart` take the "invisibility" of the `DynamicMethod` `RuntimeMethodHandle` into account. Mono holds a `private RuntimeMethodHandle mhandle`, but for now, the normal getter does the job as Mono isn't following the .NET limitation here.

Something I've noticed separately from Harmony when I implemented a runtime detouring utility is that JIT inlining can invalidate the detouring efforts... yet `[MethodImpl(MethodImplOptions.NoInlining)]` doesn't seem to cause any changes while testing.  
I'll try to fix this before finally deeming this "pullable" :)

pardeike:
Nice work there

0x0ade:
The `TestMethod1` test now passes. A small oversight caused a method to return a different method starting address when non-JITed than when JITed. I assume the debugger force-JITed the patched method. Nothing a small `RuntimeHelpers.PrepareMethod` can't fix :)

The test now takes this possibility into account (differing method start pre and post patch) and compares the method start to be completely sure (that's what actually got me on the right track).

I now deem this mergeable

jeason1997:
Nice job!! thanks

0x0ade:
I don't really like bumping and understand if you're busy. A quick review would be nice, though, so I could change anything if required :)

pardeike:
Hey,

You are totally right. I just got caught in the chaos where Tynan Sylvester and Dean Hall tweeted about my new Zombie mod for RimWorld. I got distracted and am very sorry. 

This is actually something I was looking for a long time. I get too it ASAP. Give me this week. 

Andreas Pardeike
IT-Arkitekt för Mobilitet, @pardeike
Polismyndigheten, +46 722 120 680

Kittyfisto:
@0x0ade Should this patch cause Harmony to work with .NET Framework 4.5.2?

I'm currently trying to use Harmony to overwrite the behaviour of the System.Windows.Input.Keyboard class, however patching still fails, albeit with a different exception.

In case it's already supposed to work, the following code:
`public static class TestKeyboard
	{
		private static readonly HashSet<Key> PressedKeys;

		static TestKeyboard()
		{
			PressedKeys = new HashSet<Key>();
		}

		public static void Press(Key key)
		{
			PressedKeys.Add(key);
		}

		public static void Release(Key key)
		{
			PressedKeys.Remove(key);
		}

		[HarmonyPatch(typeof(Keyboard))]
		[HarmonyPatch("IsKeyDown")]
		class PatchIsKeyDown
		{
			static void Postfix(Key key, ref bool __result)
			{
				__result = PressedKeys.Contains(key);
			}
		}
	}`
causes a System.ArgumentException when calling PatchAll.

(I can confirm, however, that your patch fixes the previous NullReferenceException due to the different method name)

0x0ade:
@Kittyfisto Good point, I'm only able to test it with the .NET Framework 4.6 on Windows 10 here. I've seen that a method signature mismatche between <= 4.5 and >= 4.6 and tried to add a few checks as required. I'm currently not at my PC, but _CompileMethod, IntPtr and internal method info ring a bell. I'll be thankful if you can provide more info to _CompileMethod on 4.5 and before.

Kittyfisto:
@0x0ade I just realized that I was looking at the .NET Version referenced by my project, not the version I've actually installed (since 4.x is replaced and not installed side-by-side).

I'm actually using 4.6.2 where the signature is as you've mentioned above, so Patch() should fail for the both of us, am I right?

Anyways, I've been playing around with the sources and made two changes that cause Patch() to no longer throw exceptions at me. I've submitted a pull request that details those changes.

pardeike:
This seems to be so mature that I’ll accept the pull request. Wouldn’t mind though if you guys would keep an eye on that part and keep improving in case of more problems. Good job!

0x0ade:
Weirdly enough, I didn't encounter the exception during testing. To be sure, I re-ran it in debug mode and...
image
... the test project ran targeting the .NET Framework 3.5! Switching to 4 triggered the same exception.

I'll take a look at your pull request on my fork now.

Kittyfisto:
Thanks. That brings the question on to how testing might be expanded to catch differences between the various major versions of the .NET Framework. I've yet to find a better solution other than to actually have multiple projects (using the same sources), targeted against various frameworks and/or platforms (x86/x64), but maybe this is a possible solution?

pardeike:
I hope I can put some more time and focus on testing during the summer. Thanks for your contribution, btw!

Kittyfisto:
You're welcome. Harmony has been the least frustrating solution in my quest to enable simple unit test of WPF controls, so there's that ;)

Issue #18, opened 2017-06-15
Patching instance methods


Kittyfisto:
I'm trying to use Harmony to overwrite the behaviour of [MouseEventArgs.GetPosition()](https://msdn.microsoft.com/en-us/library/ms591423(v=vs.110).aspx) method. However applying the following patch:

		[HarmonyPatch(typeof(MouseEventArgs))]
		[HarmonyPatch("GetPosition")]
		sealed class PatchMouseEventArgsGetPosition
		{
			private static void Postfix(MouseEventArgs __instance, IInputElement relativeTo, ref Point __result)
			{
				Positions.TryGetValue(relativeTo, out __result);
			}
		}

throws a System.NotSupportedException exception with the following message : "Wrong MethodAttributes or CallingConventions for DynamicMethod. Only public, static, standard supported".

According to [documentation](https://msdn.microsoft.com/en-us/library/0d9fd769(v=vs.110).aspx), the only supported combination is Public | Static which is why I don't understand the following code inside DynamicTools.CreateDynamicMethod:

			var method = new DynamicMethod(
				patchName,
				MethodAttributes.Public | (original.IsStatic ? MethodAttributes.Static : 0),
				CallingConventions.Standard,
				AccessTools.GetReturnedType(original),
				paramTypes,
				original.DeclaringType,
				true
			);

Is there a case (or a platform) where creating a non-static DynamicMethod is ever possible?
I've changed it locally to just MethodAttributes.Public | MethodAttributes.Static which caused my problem to go away, but I may be missing something here.

pardeike:
I would like to know in which environment you are executing Harmony. On mono, which is the main focus of this lib, a DynamicMethod is static but can emulate an instance method by choosing the correct attributes. 

In case you are executing this on Core you might want to test again with the lastest commits. I just accepted a pull request that deals with that environment.

Kittyfisto:
I'm running this code on Win10, using .NET 4.6.2 and I'm actually working with those changes you've merged earlier, including two little changes of my own to allow patching of static method under .NET 4.6.2, as detailed [here](https://github.com/0x0ade/Harmony/pull/1).

0x0ade:
From what I understand, the only required criterion to emulate an instance method properly is using the first argument (technically 0th argument) passed as `this`, very similar to how extension methods work. I haven't seen anything else inside the method body that would need changes. And also from my understanding, we're just caring about the jump target body, not its attributes.

[...]

PR #19, opened 2017-06-15
Further .NET Framework support fixes

0x0ade:
Thanks to @Kittyfisto's work, Harmony is now functional on .NET Framework 4.6.2. I didn't notice the issues during the previous pull request.

Additionally, I've added a `TestMethod2` test case that tests patching instance methods. As @Kittyfisto noticed in #18, the .NET Framework is stricter and thus requires `DynamicMethod`s to always be static.

From my understanding, Harmony only cares about the target method's body (the method to jump into), not if the target `DynamicMethod` is static or not. Furthermore, I haven't noticed any issues with static methods mimicking instance method behavior. To me, it seems quite similar to how extension methods work.


pardeike:
I agree with keeping DynamicMethod static

Issue #11, opened 2017-04-10
[Suggestion] Unpatching methods


zatherz:
I went through the wiki and some of the source code and I can't find any method for doing this. Would it be feasible to implement an interface for removing a hook from a method?

Beowulfe222:
Agreed, this would be a very useful function... If it's already in place, would great to have some information on how to handle it.

pardeike:
I will consider this in the next release unless it breaks stuff. But I am confident that it does not.

pipe01:
Any updates on this?

pardeike:
I’m stuck in West Sahara right now, so this has to wait until next month. Sorry.

pipe01:
That's cool, don't worry. Enjoy your visit...?

pardeike:
Hi again. My trip was... interesting :-)
I close this issue because I just added two API calls:
```csharp
public void RemovePatch(MethodBase original, HarmonyPatchType type, string harmonyID = null);
public void RemovePatch(MethodBase original, MethodInfo patch);
```
to `HarmonyInstance`. The first one scans for all patches matching the last two parameters (harmonyID "*" is a wildcard and null means current instance only). The second one takes a patch method and removes it.

Enjoy

pipe01:
That's awesome, thanks! But how would I get that `harmonyID`?

pardeike:
You probably do not need to. If you have the MethodInfo of your patch, just use the second form of RemovePatch(). If you want to remove some elses patch, you need to find out what ID they use to instantiate Harmony. Or you use the GetPatchInfo() function to list all patches. Each patch has the methodinfo and of course the harmony ID too.

pipe01:
Alright, thanks again!

PR #26, opened 2017-07-17
Allow undoing patches partially and completely

rumkex:
As described in #11. Since implementing it looked pretty straightforward, here you go. HarmonyInstance now has Restore method, mirroring Patch signature.
Basic tests are included.

I have a question, however: when patching the original method, what is being overwritten by the jump instruction, exactly? I took care of restoring the overwritten part, but it doesn't seem to affect the execution either way, why?

### DIFF
--- a/Harmony/HarmonyInstance.cs
+++ b/Harmony/HarmonyInstance.cs
@@ -76,6 +76,12 @@ namespace Harmony
 			processor.Patch();
 		}
 
+		public void Restore(MethodBase original, HarmonyMethod prefix, HarmonyMethod postfix, HarmonyMethod transpiler = null)
+		{
+			var processor = new PatchProcessor(this, original, prefix, postfix, transpiler);
+			processor.Restore();
+		}
+
 		//
 
 		public Patches IsPatched(MethodBase method)
--- a/Harmony/Patch.cs
+++ b/Harmony/Patch.cs
@@ -85,23 +85,46 @@ namespace Harmony
 
 		public void AddPrefix(MethodInfo patch, string owner, int priority, string[] before, string[] after)
 		{
-			var l = prefixes.ToList();
-			l.Add(new Patch(patch, prefixes.Count() + 1, owner, priority, before, after));
-			prefixes = l.ToArray();
+			AddPatch(patch, ref prefixes, owner, priority, before, after);
 		}
 
 		public void AddPostfix(MethodInfo patch, string owner, int priority, string[] before, string[] after)
 		{
-			var l = postfixes.ToList();
-			l.Add(new Patch(patch, postfixes.Count() + 1, owner, priority, before, after));
-			postfixes = l.ToArray();
+			AddPatch(patch, ref postfixes, owner, priority, before, after);
 		}
 
 		public void AddTranspiler(MethodInfo patch, string owner, int priority, string[] before, string[] after)
 		{
-			var l = transpilers.ToList();
-			l.Add(new Patch(patch, transpilers.Count() + 1, owner, priority, before, after));
-			transpilers = l.ToArray();
+			AddPatch(patch, ref transpilers, owner, priority, before, after);
+		}
+
+		private void AddPatch(MethodInfo patch, ref Patch[] patchlist, string owner, int priority, string[] before, string[] after)
+		{
+			var l = patchlist.ToList();
+			l.Add(new Patch(patch, (patchlist.LastOrDefault()?.index ?? 0) + 1, owner, priority, before, after));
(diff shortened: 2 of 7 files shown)


DRKV333:
The backup/restore part is not necessary, because Harmony does not modify the IL instructions of the original method. It places a jump at the top of the native code generated by the just-in-time (JIT) compiler during run time.

pardeike:
I just revisited this feature and I agree with DRKV333. One important thing to remember is that Harmony needs to work across different application domains. Which means that as soon as you share some data structure (beside system wide things like Lists and MethodBase) you are aiming for a potential "cannot assign PatchInfo to PatchInfo" problem since both types are virtually the same but effectively different.

I have the problem with CodeInstruction in transpilers and I jump through a lot of defensive reflections to solve it. I doubt this is worth the trouble in this case and vote strongly for the very simplictic solution to simple remove a patch and let harmony overwrite the method once again. That works fine for adding and will work fine for removing. The only thing we won't solve is the edge case of removing the last patch and returning to a pristine original. In the simple solution you end up with an assembler jump to a identical copy of the method. Which is a good compromise.

rumkex:
I see your point. I'll fix that then, unless you want to do it yourself.

pardeike:
My new implementation has the following API:
```csharp
static Main()
{
	HarmonyInstance.DEBUG = true;
	var harmonyId = "net.pardeike.test.harmony";

	var instance = HarmonyInstance.Create(harmonyId);
	instance.PatchAll(typeof(Main).Assembly);

	// NEW: call RemovePatch with patch id, method and patch type
	var m_PreClose = AccessTools.Method(typeof(Window), "PreClose");
	instance.RemovePatch(harmonyId, m_PreClose, HarmonyPatchType.Postfix);
}
```

pardeike:
If this is ok, I'll close this pull request. I appreciate the effort and discussion nevertheless!

rumkex:
Of course. Go ahead, it's the feature that matters. Does it handle the case where there are multiple patches of the same type, though? Unless that's impossible in Harmony?

pardeike:
Would love to suppor multiple patches within the same harmony id scenario. Problem is, that would exclude attribute patching since you never get a reference for the patch operation and it would be hard to build it in. How edge casey is that you patch the same method more than once in your implementation?

pardeike:
For
```public void Patch(MethodBase original, HarmonyMethod prefix, HarmonyMethod postfix, HarmonyMethod transpiler = null);```
I could return the reference of the original patch but since that could potentially be copied by another assembly running Harmony, it is rather tricky to implement correctly:

```public PatchReference Patch(MethodBase original, HarmonyMethod prefix, HarmonyMethod postfix, HarmonyMethod transpiler = null);```

I'll think about it.

**Comment by pardeike** [comment]
The only way to solve this is to give every patch a user-chosen id so you can refer to it later. But that might break backwards compatibility with mixed Harmony version scenarios. Rather want to avoid this for now.

**Comment by rumkex** [comment]
> How edge casey is that you patch the same method more than once in your implementation?

The code that I've written is usually fairly simple, but I can imagine complex projects having to compartmentalize patches (unless you want to have a god-hook with ALL the handling going on in it)

And in case of your original implementation, once you _do_ add multiple patches, it's going to be pretty much undefined behavior for the patch removal (removing the first patch in the list? playing Russian roulette with patches?)

What if you accepted MethodInfo for the patch itself in addition to (or instead of) original MethodInfo?

**Comment by pardeike** [comment]
Yes, of course. Just after I wrote my comment I realized that too. Going to add that as an overwrite to my current API. Like this:

```csharp
public void RemovePatch(string harmonyID, MethodBase original, HarmonyPatchType type);
public void RemovePatch(string harmonyID, MethodBase original, MethodBase patch);
```

First one removes anything, second one is selective.

pardeike:
Also, I am going to make the id optional, since it is already stored in the instance but will keep it so you can remove others patches too.

rumkex:
> Also, I am going to make the id optional, since it is already stored in the instance but will keep it so you can remove others patches too.

This is going to open a whole new can of worms since people won't generally expect that another mod can remove their patches at will.

pardeike:
You can at any time overwrite a method with a transpiler to do anything to your liking anyway (or a prefix returning false). There is no way to enforce niceness. But in case you want to go smart and, in a sensitive and responsible way, remove a different mods patch, you can. I am not going to interfer with that.

rumkex:
I see, fair enough.

pardeike:
I have actively started developing on Harmony. I am going to close this now and will release updates in a regular fashion now. Thanks for your patience.

PR #48, opened 2017-10-06
Added HarmonyPatchAll and HarmonyMethodFilter Attributes.

FrodoOf9Fingers:
Usages:

```
[HarmonyPatchAll]
class RandomPatch
{
        [HarmonyPrefix]
        public static bool prefixAll(PathFinder __instance, object __state)
        { ...}
        [HarmonyPostFix]
        public static bool postfixAll(PathFinder __instance, object __state)        
        {...}
}

[HarmonyPatchAll]
[HarmonyMethodFilter(BindingFlags.Instance | BindingFlags.Public)]
class AnotherRandomPatch
{...}
```

I did not implement a Type[] parameter, as it seemed that if you needed the parameters of the function you were accessing you should be implementing individual patches. __instance and __state work, but __result requires an explicit type and has been left unsupported. Finally, transpilers defined in a HarmonyPatchAll will throw an exception.

### DIFF
--- a/Harmony/Attributes.cs
+++ b/Harmony/Attributes.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Reflection;
 
 namespace Harmony
 {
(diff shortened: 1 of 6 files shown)


pardeike:
HarmonyPatchAll is now implemented. Filtering is done by using the existing `bool Prepare(MethodBase method)` or by emitting the correct list of methods to patch in `IEnumerable<MethodBase> TargetMethods()`.

PR #52, opened 2017-11-29
Added ability to give parameters your own name

### LINKED ISSUE #51: Obfuscated parameters
What if the method has obfuscated parameters that begin with # ? 

How could I access them?

Also how does it deal with multiple overloads?

atomicwrangler:
Fixes #51 

First I wanted to add a new field to PatchInfo/Patch but I changed my mind. If you have any design/optimization suggestions or you have any objections regarding current design tell me. Feel free to change the naming as well. I tried to follow your naming convention.

```csharp
[HarmonyPatch(typeof(Player))]
[HarmonyPatch("CastSpell")]
[HarmonyParameter("#d", "spellId")]
class ExamplePatch
{
    // This has the higher priority than the one above
    [HarmonyParameter("#d", "spellId")]
    [HarmonyParameter("#c", "targetid")]
    // Parameter attributes have the highest priority
    public static void Prefix([HarmonyParameter("#d")] int spellId, int targetid)
    {
    }
    
    // No need to do anything since we have added class-wide attribute
    public static void Postfix(int spellId)
    {
    }
}
```

This also works when you use `Harmony.Patch` instead of `Harmony.PatchAll`

### DIFF
--- a/Harmony/Attributes.cs
+++ b/Harmony/Attributes.cs
@@ -116,17 +116,17 @@ namespace Harmony
 	[AttributeUsage(AttributeTargets.Parameter | AttributeTargets.Method | AttributeTargets.Class, AllowMultiple = true)]
 	public class HarmonyParameter : Attribute
 	{
-		public string Name { get; private set; }
-		public string CustomName { get; private set; }
+		public string OriginalName { get; private set; }
+		public string NewName { get; private set; }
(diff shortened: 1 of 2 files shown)


atomicwrangler:
Please check if it's fine now. I renamed them.

PR #73, opened 2018-04-03
Converted to VS 2017 project, which automatically builds a NuGet package

wasabii:
@pardeike I wanted to get a NuGet package for this, so I could deploy it to my own private NuGet feed to use inside my company.

Your project was so super small, that I figured it'd be easy to just convert it to the new VS2017 project type, which does automatic NuGet packages. It was. So I did.

Here are the results.


pardeike:
Sounds great. I’ll look into it tonight.

pardeike:
Question: don't I loose the ability to generate just the `0Harmony.dll` for local deployment (or if someone is not using VS because they are on Linux/Mono)? I would love to have both: a nuget publishable package and an ordinary dll.

wasabii:
https://github.com/dotnet/cli/issues/5723

mhutch indicates that the new 2017 project style should have been working in early 2017 on Mono.

HoLLy-HaCKeR:
The new csproj format will always place the original DLL files in a folder, alongside an optional nuget package. You also get to build for multiple target frameworks (for example .NET Standard 2.0+.NET 3.5). The downside is that you need VS2017 and that nuget packages aren't stored in the project's directory anymore.

Your files will be places in folders like this:
```
./bin/release/netstandard20/0Harmony.dll
./bin/release/net4.7.1/0Harmony.dll
./bin/release/Harmony.nuget
```

Issue #20, opened 2017-06-18
Discussion on a new field injection feature


pardeike:
Hi,

I am thinking of adding more argument injections to patch methods. Currently, we have
- original method arguments (by exact name and type)
- **__instance** to get to *this* for non-static methods
- **__state** to create a local variable holding state between prefix and postfix
- **__result** to access the original methods result value

What often happens is that you want to access private fields of the original methods class but it adds a lot of overhead in form of performance and code to your patch methods.

So my proposal is that you can add arguments to the patch methods that match the type and name of a field and Harmony will inject those for you.

Example:

```csharp
class Foo
{
	private List<String> aList;
	
	private void SomeMethod(String foo)
	{
		aList.Add(foo);
		// ...
	}
}
```

could be patched with

```csharp
[HarmonyPatch(typeof(Foo))]
[HarmonyPatch("SomeMethod")]
static class FooPatch
{
	static void Prefix(ref String foo, List<String> aList)
	{
		// ...
	}
}
```

Some reflections:

- Naming could be an issue. I don't want to overcomplicate things with a required name prefix and I assume giving the original method argument names that are the same as its fields will cause conflicts anyway, or are there edge cases?
- Are there other cases of injections that are useful?

Anyone care to discuss this?

DoctorVanGogh:
Stupid idea time :grin:

## Get rid of the hardcoded parameter names (keep it as base case, but offer an indirection layer):
```csharp
enum SpecialParam {
  Result,
  Instance,
  State,
}

[AttributeTarget(AttributeTargets.Parameter, AllowMultiple=false)]
class ParamAttribute : Attribute {
  ParamAttribute (SpecialParam type) {
    ...
  }
  ParamAttribute (string parameterName) {
    ...
  }
}
[AttributeTarget(AttributeTargets.Parameter, AllowMultiple=false)]
class FieldAttribute : Attribute {
  FieldAttribute (string fieldName) {
...
 }
}

```
Then consumers could simply declare

```csharp
[HarmonyPatch(typeof(Foo), nameof(Foo.Thingy)]
class Foo_ThingyPatch {
    static void Postfix(
        [Param(SpecialParam.Result)] List<Bar> parameterExplicitlyNotNamedAnythingLikeResult, 
        [Param(SpecialParam.Instance)] Foo foo,
        [Param("methodParam1")] int p1,
        [Field("privateFieldXYZ")] Baz xyz
    ) {
    }
}

```
[...]

PR #93, opened 2018-05-23
Add convenient accessor class

enemyofthedawn:
Something what was proposed here: https://github.com/pardeike/Harmony/issues/20#issuecomment-311988446

Basically it's a nice looking helper(under the hood using`Traverse`) for accessing/modifying field and property values. 

Example:

```csharp
var accessor = Accessor.ForInstance(someinstance);
var fieldValue = accessor.Field<int>("_someIntField").Value;
var propertyValue = accessor.Property<int>("SomeIntProperty").Value;

var healthFieldReference = accessor.Field<int>("_health");
// Modifying values is possible too
healthFieldReference.Value = 999;

```

### DIFF
--- a/Harmony/Tools/Accessor.cs
+++ b/Harmony/Tools/Accessor.cs
@@ -1,0 +1,57 @@
+using Harmony;
+using System;
+
+namespace Harmony
+{
+	public class Reference<T>
+	{
+		private Traverse traverse;
+
+		public Reference(Traverse traverse)
+		{
+			this.traverse = traverse;
+		}
+
+		public T Value
+		{
+			get { return traverse.GetValue<T>(); }
+			set { traverse.SetValue(value); }
+		}
+	}
+
+	public class Accessor
+	{
+		private Traverse traverse;
+
+		public Accessor(object instance)
+		{
+			traverse = new Traverse(instance);
+		}
+
+		public Accessor(Type type)
+		{
+			traverse = new Traverse(type);
+		}
+
+		public static Accessor ForInstance(object instance)
+		{
+			return new Accessor(instance);
+		}
+
+		public static Accessor ForType(Type type)
+		{
+			return new Accessor(type);
+		}
+
+		public Reference<T> Field<T>(string fieldName)
+		{
+			return new Reference<T>(traverse.Field(fieldName));
+		}
+
+		public Reference<T> Property<T>(string propertyName)
+		{
+			return new Reference<T>(traverse.Property(propertyName));
+		}
+	}
+
+}


pardeike:
Is the sole benefit of this API to allow writing
```csharp
var accessor = Accessor.ForInstance(someinstance);
var fieldValue = accessor.Field<int>("_someIntField").Value;
```
instead of
```sharp
var fieldValue = Traverse.Create(someinstance).Field("_someIntField").GetValue<int>();
```
or
```sharp
var traverse = Traverse.Create(someinstance);
var fieldValue = traverse.Field("_someIntField").GetValue<int>();
```
or
```sharp
var traverse = Traverse.Create(someinstance).Field("_someIntField");
var fieldValue = traverse.GetValue<int>();
```
I am not sure if that adds a sufficient benefit compared to the complexity to document and explain the class to end users. Too many choices can be contra productive.

What is the motivation behind your proposal?

pardeike:
Also note, that we already have extra API like FastAccess:
https://github.com/pardeike/Harmony/blob/master/Harmony/Extras/FastAccess.cs

fromdusktilldawn:
Under the hood it still uses `Traverse` but it's easier to write and looks cleaner. So it can be treated more like a wrapper for `Traverse`. I created it because at some point I started using `Traverse` and it was really annoying to use. I didn't notice there's something like FastAccess, but still it's not that convenient to use. Maybe it would be faster to use `FastAccess` instead of `Traverse` for this class.

Personally I think it is much better to assign values like:

`someField.Value = 1` than `someField.SetValue<int>(1)`

pardeike:
I am not going to accept this pull request. But I am going to try to incorporate your simplified API into Traverse directly. Your class is too similar to Traverse and does not offer chaining. So by instead adding the core simplification into Traverse, we get the best of both worlds.

If we add your `Value getter/setter` to traverse we get the same result, don't we? Or am I missing something here?

fromdusktilldawn:
And how are you going to get `T`? You can't have generic property in C#, `Traverse` class is not generic and this kind of syntax won't be possible.

pardeike:
I understand where you are coming from. I just think keeping most of Traverse and its feature and adding your final step to it gives the same result without a whole new way to define type and field names.

So my proposal is to add
```csharp
public class Accessor<T>
{
	private Traverse traverse;

	public Accessor(Traverse traverse)
	{
		this.traverse = traverse;
	}

	public T Value
	{
		get { return traverse.GetValue<T>(); }
		set { traverse.SetValue(value); }
	}
}
```
and
```csharp
public Accessor<T> Field<T>(string name)
{
	return new Accessor<T>(Field(name));
}

public Accessor<T> Property<T>(string name, object[] index = null)
{
	return new Accessor<T>(Property(name, index));
}
```
directly to `Traverse`. Which allows to write:
```csharp
var someField = Traverse.Create(someInstance).Field<FieldType>("someField");
someField.Value = ...;
```

PR #95, opened 2018-06-21
Fix issue formatting ldstr operand

amPerl:
Previously if you were to call ToString on a `ldstr` instruction with a constant string that included format indices like `{1}`, then it would get passed directly to `string.Format`, causing an exception.

This PR removes the seemingly unnecessary `string.Format` call (as no other parameters are passed) and includes a test to cover `ToString()` exceptions.


pardeike:
Awesome!

PR #104, opened 2018-07-10
Nuget package fixes

wasabii:
Remove independent configuration for Release and Debug versions of 3.5.
Update HarmonyTests to latest package format: add MsTest. Had to remove 3.5 from this project, as new version of MSTest isn't compatible with 3.5 any longer. 4.5 tests still run. I'm totally not sure if this matters?


wasabii:
Oh, yeah. That's how did usually do it

pardeike:
Just checked. In order to publish on nuget you need a signing certificate that is produced by a microsoft known CA (Symantec etc). These cost money so I rather not want to go that way. Also, Microsoft states that self signed certificates are [not accepted](https://docs.microsoft.com/en-us/nuget/reference/signed-packages-reference).

PR #109, opened 2018-07-18
add simple constructor-targetting overload for patch attribute

DoctorVanGogh:
attribute refactoring ate the `Type, Type[]` overload


pardeike:
Was added with my other changes, so it's obsolete.

Issue #113, opened 2018-08-06
Request: Reference parameters and global variables by index


FailedShack:
It would be really convenient if parameters and local variables could be referenced by their index as well as their name in patch methods as obfuscated assemblies sometimes use very long nonsensical field names and usually vary between different versions.

For referencing parameters, this could be done by naming the patch method parameters like __0, __1, __2, where the number prepended by two underscores represents the index at which the original parameter is located.

For global variables it would be the same thing but instead of being prepended by two underscores it would be prepended by three instead.

Since the C# specification doesn't support field names that start with digits, this shouldn't clash with anything else.

With a target method looking like this:
![target](https://user-images.githubusercontent.com/39192928/43725370-a26eb136-999c-11e8-9c28-d3c9c44bec77.PNG)

You would end up with a patch method looking like this:
![patch](https://user-images.githubusercontent.com/39192928/43725493-f6052f6e-999c-11e8-80ec-1a68b3e9964c.PNG)

Instead of like this:
![patch1](https://user-images.githubusercontent.com/39192928/43725612-4e23145e-999d-11e8-841f-51c84cc98f68.PNG)

pardeike:
Can you explain how local var injection would work in principle?

pardeike:
I like the __n form of addressing arguments though. 

And if I am not mistaken there is already an index annotation and a rename annotation for arguments.

FailedShack:
@pardeike Sorry, I mistakenly wrote 'local' instead of 'global', already edited.

Just found it in the Attributes class, thought it didn't exist because it's not in the Wiki.
Before your reply I had been toying with the solution proposed in this issue and quickly implemented it in https://github.com/FailedShack/Harmony/commit/8bad1dec173c784b91ba09338aaf56a75b7b373c

Since you can simply rename it I assume this is not really necessary but I think it would be cool if you could directly reference the indexes as well, so I could open a pull request if you wish.

pardeike:
By “global variables” you mean fields in the class. I’ll commented on your pull request and will accept it if you polish it a bit.

PR #114, opened 2018-08-06
Support for directly addressing parameters and fields by index

FailedShack:
See #113


pardeike:
Looks great. Would you mind adapting the error messages in both cases a bit? The first one would say “field __3” which isn’t helpful and the second case needs a bit of error reporting if the index is out of valid range.

pardeike:
Excellent.

PR #127, opened 2018-09-08
Improved Exception Throwing from HarmonyMethod & MethodPatcher

AaronCRobinson:
- Do not allow HarmonyMethod to refer to a non-existant method
- Do not allow Postfix patch to change return type (bad ducks)

### DIFF
--- a/Harmony/MethodPatcher.cs
+++ b/Harmony/MethodPatcher.cs
@@ -328,6 +328,8 @@ namespace Harmony
 				{
 					if (AccessTools.GetReturnedType(original) == typeof(void))
 						throw new Exception("Cannot get result from void method " + original.FullDescription());
+					if (AccessTools.GetReturnedType(original) != patchParam.ParameterType)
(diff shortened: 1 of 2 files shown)


Harmony/MethodPatcher.cs
```diff
@@ -328,6 +328,8 @@ static void EmitCallParameter(ILGenerator il, MethodBase original, MethodInfo pa
 				{
 					if (AccessTools.GetReturnedType(original) == typeof(void))
 						throw new Exception("Cannot get result from void method " + original.FullDescription());
+					if (AccessTools.GetReturnedType(original) != patchParam.ParameterType)
```
pardeike: I don't think types should be compared with == here. What is needed here is a check for assignable since the patch could have a superclass or even `object` as a return type.

pardeike:
Fixed in 6cb008dbdf0ba78a52e8294d8e647316c236f98b

PR #133, opened 2018-10-18
Fix spelling of "priority" field name

ayan4m1:
The `priority` field in `HarmonyMethod` is currently misspelled `prioritiy`. This typo also occurs in the parameter name for the explicit `HarmonyPriority` constructor.

[This line](https://github.com/pardeike/Harmony/compare/master...ayan4m1:fix/priority-name?expand=1#diff-61f938e2025ec01f805d23e93979ad2cR17) may be a breaking change for downstream users, but the fix is dead simple.

Also, apologies as I am not intimately familiar with the codebase, but is there a reason why the public fields of `HarmonyMethod` are camelCase and not CamelCase?

### DIFF
--- a/Harmony/Attributes.cs
+++ b/Harmony/Attributes.cs
@@ -218,9 +218,9 @@ namespace Harmony
 	[AttributeUsage(AttributeTargets.Class | AttributeTargets.Method)]
 	public class HarmonyPriority : HarmonyAttribute
 	{
-		public HarmonyPriority(int prioritiy)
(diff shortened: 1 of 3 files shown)


pardeike:
I hardly see any breaking change in this. Good catch!
In general, I keep the style VS recommends. Which is methods are CamelCase and fields are camelCase. At least I try to be consistent with that.

ayan4m1:
Cool, thanks for the info!

PR #140, opened 2018-11-26
slightly improved error reporting

Mehni:
Ambiguous matches that were caught now throw when there's no match and inform you where you screwed up.

Null originals now show the Harmony.ID rather than just the vague error of "original". Should people screw up their overloads or typo the target method, they can theoretically change their harmony ID to more easily find the offending patch.


pardeike:
Thanks!

PR #141, opened 2018-11-26
HarmonyAfter Issue

catgirlfighter:
There must be a feedback. In ordinary sorting, when x is compared to y, y does not get compared to x afterwards (what should be obvious logically). That's why HarmonyAfter of a patch higher in the list never gets applied.

### DIFF
--- a/Harmony/Patch.cs
+++ b/Harmony/Patch.cs
@@ -49,19 +49,25 @@ namespace Harmony
 		}
 
 		// general sorting by (in that order): before, after, priority and index
-		public static int PriorityComparer(object obj, int index, int priority, string[] before, string[] after)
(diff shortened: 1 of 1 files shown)


pardeike:
I revisited this request and it turns out to be much more complicated. Given three patches with equal priority and added in this order:

Patch1, after:Patch3
Patch2
Patch3

The comparer will compare 1+2 (result 1<2), then 2+3 (result 2<3) and stop. This does not give the expected result of 2, 3, 1 even with your code changes. A different compare algorithm is needed.

catgirlfighter:
Yes, I agree with your decision. It worked in my particular case, but looking deeper into the case I noticed that it would require a lot more to properly address all possibilities.

pardeike:
Solved by PR #155

PR #142, opened 2018-11-28
further improvements to error reporting

Mehni:
similar to my previous PR. Should be self-explanatory.

Some thoughts: 
- It doesn't throw but maybe it should. As far as I can tell, MethodPatcher.CreatePatchedMethod() is where it should throw, but by then it's too late in a lot of regular use cases as Harmony will likely have already encountered a NRE elsewhere. Throwing this early may have side-effects I'm not aware of.

- AccessTools.DeclaredMethod(), which is very similar in structure, does not have the same level of try/catch and error reporting.


pardeike:
I took the liberty to extend the concept to all of AccessTools in commit 23957d1b816325ce44c6751f38cb220ea3a82ab2

PR #146, opened 2019-01-02
Avoid System.ArgumentException

DenJur:
Avoids potential System.ArgumentException when null method is added to the cache multiple times due to AccessTools failing to find method that fits the signature.

Not sure if that's the exact behavior you want to have or if throwing exception would be better.


pardeike:
Looks good

pardeike:
I think throwing would better be placed somewhere higher up with a context error

PR #153, opened 2019-01-29
Use generics for FastAccess

DenJur:


DenJur:
This is why you shouldn't code when you have been awake for 50 hours 😄 
Should be fixed now

PR #160, opened 2019-02-26
FieldRefAccess expansion

DenJur:
Added overload that takes FieldInfo as an argument. Also added support for static fields.
Not sure what behavior you want for a version that takes in a fieldname. For now I left it as it was in the original.

### DIFF
--- a/Harmony/Tools/AccessTools.cs
+++ b/Harmony/Tools/AccessTools.cs
@@ -733,14 +733,27 @@ namespace Harmony
 											BindingFlags.DeclaredOnly;
 
 			var fi = typeof(T).GetField(fieldName, bf);
-			if (fi == null)
-				throw new MissingFieldException(typeof(T).Name, fieldName);
+			return FieldRefAccess<T,U>(fi);
+		}
 
-			var s_name = "__refget_" + typeof(T).Name + "_fi_" + fi.Name;
+		/// <summary>Creates a field reference</summary>
+		/// <typeparam name="T">The class the field is defined in</typeparam>
+		/// <typeparam name="U">The type of the field</typeparam>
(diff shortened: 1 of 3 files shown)


Harmony/Tools/AccessTools.cs
```diff
@@ -733,23 +733,43 @@ public static object[] ActualParameters(MethodBase method, object[] inputs)
 											BindingFlags.DeclaredOnly;
 
 			var fi = typeof(T).GetField(fieldName, bf);
-			if (fi == null)
-				throw new MissingFieldException(typeof(T).Name, fieldName);
```
pardeike: Why is this being removed?

HarmonyTests/Tools/TestAccessTools.cs
```diff
@@ -169,5 +169,80 @@ public void AccessTools_TypeExtension_Types()
...
+	  }
+
+	  [TestMethod]
+		public void AccessTools_FieldRefAccess_ByFieldInfo_Static()
+		{
+			var fieldInfo = typeof(AccessToolsClass).GetField("field3", BindingFlags.Static | BindingFlags.NonPublic);
+			Assert.IsNotNull(fieldInfo);
+			var instance = new AccessToolsClass();
+			var fieldRef = AccessTools.FieldRefAccess<AccessToolsClass, String>(fieldInfo);
+			ref String value = ref fieldRef(instance);
```
pardeike: Passing an instance makes no sense for a static field

Harmony/Tools/AccessTools.cs
DenJur: It is checked later

pardeike:
See my comments. Maybe an overload for static fields would be better?

PR #174, opened 2019-04-05
Doc fix + ArgumentException on null or empty id

Mehni:
b3368e3 fixes a tabs vs spaces mistake I introduced in #167.

The argument exception we discussed on discord.


Pathoschild:
Maybe `string.IsNullOrWhiteSpace` would be better, unless whitespace-only IDs are fine?

Mehni:
It would, but that doesn't exist in 3.5.

PR #182, opened 2019-05-01
Fixes for DynamicMethod patching

bbepis:
I've also included a unit test which is what I've been using for reproduction.

Note that in `Assets.Class11` I've included an `int dummy` parameter, which is required to trigger failures that are fixed in MethodPatcher.cs line 234 and line 300.

I've tested this in both regular .NET 4.5 runtime and Unity 5.x which contains the old mono implementation. Fix on line 234 was only needed for the old mono runtime, it works fine without it on regular .NET.

I have not tested this on Unity 2017.3, however I believe the new mono runtime included with it would work, similar to the (more compliant) .NET implementation.  Unity 2018 and above do not apply as they ship mscorlib with the stubbed System.Reflection.Emit runtime.

### DIFF
--- a/Harmony/Internal/MethodPatcher.cs
+++ b/Harmony/Internal/MethodPatcher.cs
@@ -60,7 +60,7 @@ namespace HarmonyLib
 
 				prefixes.Union(postfixes).Union(finalizers).ToList().ForEach(fix =>
 				{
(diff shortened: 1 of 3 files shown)


pardeike:
Nice work!

PR #183, opened 2019-05-04
fix bug, Exception Filter block should call BeginCatchBlock 

stonewell:
to make ILGenerator stay in correct state

### DIFF
--- a/Harmony/Internal/MethodCopier.cs
+++ b/Harmony/Internal/MethodCopier.cs
@@ -162,8 +162,8 @@ namespace HarmonyLib
 				var instr2 = GetInstruction(handler_end, true);
 				instr2.blocks.Add(new ExceptionBlock(ExceptionBlockType.EndExceptionBlock, null));
 
-				// The FilterOffset property is meaningful only for Filter clauses. 
-				// The CatchType property is not meaningful for Filter or Finally clauses. 
+				// The FilterOffset property is meaningful only for Filter clauses.
+				// The CatchType property is not meaningful for Filter or Finally clauses.
 				//
 				switch (exception.Flags)
 				{
@@ -319,11 +319,19 @@ namespace HarmonyLib
 				//	if (codeInstruction.blocks.Any(block => block.blockType == ExceptionBlockType.EndExceptionBlock))
 				//		emitCode = false;
 
-				//	// skip LEAVE on next instruction starts a new exception handler and we are already in 
+				//	// skip LEAVE on next instruction starts a new exception handler and we are already in
 				//	if (idx < instructions.Length - 1)
 				//		if (instructions[idx + 1].blocks.Any(block => block.blockType != ExceptionBlockType.EndExceptionBlock))
 				//			emitCode = false;
 				//}
+
+                //for end filter exception instruction
+                //let BeginCatchBlock(null) call to emit the instruction
+                //and the API call will make ILGenerator stay in correct state
+                if (code == OpCodes.Endfilter) {
+                    emitCode = false;
+                    generator.BeginCatchBlock(null);
+                }
 
 				if (emitCode)
 				{
--- a/HarmonyTests/TestExceptionFilterBlock.cs
+++ b/HarmonyTests/TestExceptionFilterBlock.cs
@@ -1,0 +1,63 @@
+using HarmonyLib;
+using HarmonyLibTests.Assets;
+using NUnit.Framework;
+
+using System;
+using System.Collections.Generic;
+using System.Reflection;
+using System.Reflection.Emit;
+using System.Runtime.CompilerServices;
+
+namespace HarmonyLibTests
+{
+	public class ClassExceptionFilter
+	{
+		[MethodImpl(MethodImplOptions.NoInlining)]
+		public string Method0()
+		{
+            try {
+                throw new Exception("123456");
+            } catch(Exception e) when(e.Message == "123456") {
+            }
+
+            return "original";
+		}
+	}
+
+    public class ClassExceptionFilterPatch
+	{
+		public static void Postfix(ref string __result)
+		{
+			__result = "patched_exception_filter";
+		}
+	}
+
+	[TestFixture]
+	public class TestExceptionFilterBlock
+	{
+		[Test]
+		public void TestMethodExceptionFilterBlock()
+		{
+            Harmony.DEBUG = true;
+			var originalClass = typeof(ClassExceptionFilter);
+			Assert.IsNotNull(originalClass);
+			var originalMethod = originalClass.GetMethod("Method0");
+			Assert.IsNotNull(originalMethod);
+
+			var patchClass = typeof(ClassExceptionFilterPatch);
+			var postfix = patchClass.GetMethod("Postfix");
+			Assert.IsNotNull(postfix);
+
+			var instance = new Harmony("test");
+			Assert.IsNotNull(instance);
+
+			var patcher = new PatchProcessor(instance, new List<MethodBase> { originalMethod }, null, new HarmonyMethod(postfix), null);
+			Assert.IsNotNull(patcher);
+			patcher.Patch();
+
+			var result = new ClassExceptionFilter().Method0();
+			Assert.AreEqual("patched_exception_filter", result);
+		}
+
+	}
+}


pardeike:
While this is technically correct, it is not how Harmony is designed. To get the correct result in a transpiler, the user is responsible to add a `ExceptionBlockType.BeginCatchBlock` to `instruction.blocks`. 

Harmony does add that correctly to codeinstructions before any transpiler is called by looking at the exception handling structure of the original method. Any subsequent manipulation of the code instructions need to deal with that too. In Harmony, exception handling is not tied to specific codes but to a meta structure in the `blocks` field in a `CodeInstruction` and the responsibility for correctness is to each transpiler implementation.

stonewell:
Thanks for the reply. I am a little bit confuse, since the problem is happend in MethodCopier.cs, I suppose it use to copy the original method to patched method and User has no control to add a ExceptionBlock. 
Do I misunderstand How to use Harmony?

stonewell:
and I meet this problem when I try to patch a method use blow syntax
catch (HttpRequestException e) when (e.Message.Contains("301"))

pardeike:
If your error happens when you are the only one patching a method and if you are not using a transpiler, then it is a bug. 

If any transpiler changes the method (yours or someone elses, before or after you) then this problem is most likely a user error. 

I need more details and a full stacktrace plus (if possible) the original method. Also, please set 

HarmonyInstance.DEBUG = true

before you create your harmony instance and submit the resulting harmony log file from your computers desktop.

stonewell:
Microsoft (R) Build Engine version 16.1.54-preview+gd004974104 for .NET Core
Copyright (C) Microsoft Corporation. All rights reserved.

  Restore completed in 44.56 ms for /Users/stone/Work/GitHub/Harmony.origin/Harmony/Harmony.csproj.
  Restore completed in 44.82 ms for /Users/stone/Work/GitHub/Harmony.origin/HarmonyTests/HarmonyTests.csproj.
Test run for /Users/stone/Work/GitHub/Harmony.origin/HarmonyTests/bin/Debug/net45/HarmonyTests.dll(.NETFramework,Version=v4.5)
Microsoft (R) Test Execution Command Line Tool Version 16.0.1
Copyright (c) Microsoft Corporation.  All rights reserved.

Starting test execution, please wait...
Failed   TestMethodExceptionFilterBlock
Error Message:
 System.Exception : Exception from HarmonyInstance "test" patching HarmonyLibTests.ClassExceptionFilter.Method0()
  ----> System.InvalidOperationException : Incorrect code generation for exception block.
Stack Trace:
  at HarmonyLib.MethodPatcher.CreatePatchedMethod (System.Reflection.MethodBase original, System.String harmonyInstanceID, System.Collections.Generic.List`1[T] prefixes, System.Collections.Generic.List`1[T] postfixes, System.Collections.Generic.List`1[T] transpilers, System.Collections.Generic.List`1[T] finalizers) [0x005c7] in <2751aaa9396f4a8aab4810e5c2b835dc>:0
  at HarmonyLib.PatchFunctions.UpdateWrapper (System.Reflection.MethodBase original, HarmonyLib.PatchInfo patchInfo, System.String instanceID) [0x00035] in <2751aaa9396f4a8aab4810e5c2b835dc>:0
  at HarmonyLib.PatchProcessor.Patch () [0x0010d] in <2751aaa9396f4a8aab4810e5c2b835dc>:0
  at HarmonyLibTests.TestExceptionFilterBlock.TestMethodExceptionFilterBlock () [0x00085] in <d91c41ef97b143a7a0dd6ef8663896d2>:0
  at (wrapper managed-to-native) System.Reflection.MonoMethod.InternalInvoke(System.Reflection.MonoMethod,object,object[],System.Exception&)
  at System.Reflection.MonoMethod.Invoke (System.Object obj, System.Reflection.BindingFlags invokeAttr, System.Reflection.Binder binder, System.Object[] parameters, System.Globalization.CultureInfo culture) [0x0003b] in <98fac219bd4e453693d76fda7bd96ab0>:0
--InvalidOperationException
  at System.Reflection.Emit.ILGenerator.EndExceptionBlock () [0x00046] in <98fac219bd4e453693d76fda7bd96ab0>:0
  at HarmonyLib.Emitter.MarkBlockAfter (System.Reflection.Emit.ILGenerator il, HarmonyLib.ExceptionBlock block) [0x0003d] in <2751aaa9396f4a8aab4810e5c2b835dc>:0
  at HarmonyLib.MethodBodyReader+<>c__DisplayClass19_0.<FinalizeILCodes>b__5 (HarmonyLib.ExceptionBlock block) [0x00000] in <2751aaa9396f4a8aab4810e5c2b835dc>:0
  at HarmonyLib.CollectionExtensions.Do[T] (System.Collections.Generic.IEnumerable`1[T] sequence, System.Action`1[T] action) [0x0001b] in <2751aaa9396f4a8aab4810e5c2b835dc>:0
  at HarmonyLib.MethodBodyReader+<>c__DisplayClass19_0.<FinalizeILCodes>b__2 (HarmonyLib.CodeInstruction codeInstruction) [0x00256] in <2751aaa9396f4a8aab4810e5c2b835dc>:0
  at HarmonyLib.CollectionExtensions.Do[T] (System.Collections.Generic.IEnumerable`1[T] sequence, System.Action`1[T] action) [0x0001b] in <2751aaa9396f4a8aab4810e5c2b835dc>:0
  at HarmonyLib.MethodBodyReader.FinalizeILCodes (System.Collections.Generic.List`1[T] transpilers, System.Collections.Generic.List`1[T] endLabels) [0x00242] in <2751aaa9396f4a8aab4810e5c2b835dc>:0
  at HarmonyLib.MethodCopier.Finalize (System.Collections.Generic.List`1[T] endLabels) [0x00001] in <2751aaa9396f4a8aab4810e5c2b835dc>:0
  at HarmonyLib.MethodPatcher.CreatePatchedMethod (System.Reflection.MethodBase original, System.String harmonyInstanceID, System.Collections.Generic.List`1[T] prefixes, System.Collections.Generic.List`1[T] postfixes, System.Collections.Generic.List`1[T] transpilers, System.Collections.Generic.List`1[T] finalizers) [0x0023d] in <2751aaa9396f4a8aab4810e5c2b835dc>:0

Total tests: 94. Passed: 93. Failed: 1. Skipped: 0.
Test Run Failed.
Test execution time: 3.2114 Seconds

stonewell:
[harmony.log.txt](https://github.com/pardeike/Harmony/files/3145026/harmony.log.txt)

the log file, and I also submit the TestCase file in my pull request

stonewell:
[TestExceptionFilterBlock.cs.zip](https://github.com/pardeike/Harmony/files/3145032/TestExceptionFilterBlock.cs.zip)
Attached TestCase file

pardeike:
Ah, thanks for the test case. That makes a lot of sense. I will replicate the test failure and find a way to generate the correct population for the blocks fields so the exception handling code will create the call that your first commit does on the code itself.

The reason why I will not pull the contents of your first commit is that it hides the content of the filter boundary from transpilers. When reading the original message body, the exception block boundaries are parsed and added to the blocks fields, then later after all transpilers ran, taken again from the blocks field and re-emitted. Your change does not do that.

The real fix is most likely in:
https://github.com/pardeike/Harmony/blob/ca62d7e8cfef91ec64265189a1601d2a9d424e59/Harmony/Internal/MethodCopier.cs#L149
which should later insert the correct call here:
https://github.com/pardeike/Harmony/blob/ca62d7e8cfef91ec64265189a1601d2a9d424e59/Harmony/Internal/MethodCopier.cs#L293

I’ll get back to you with some feedback but for now, I cannot pull this.

pardeike:
Unless I am totally wrong about the EndFilter opcode and the exception structure does not contain the correct boundary information for this case. Which would be really strange because that opcode would exist in the original method but should not be emitted in the copy and only exist to generate the exception boundary which in itself is normally stored in the exception meta information with all the other exception boundary types.

stonewell:
I think you are correct about the logic of program, my pull is mostly "make it work" fix not logically fixing the problem.
I should make a issue instead a pull request, I am closing this pull and hopefully you could come out a real fix later, the IL structure and language stuff is too much for me:)

pardeike:
I think this is a wont-fix. See https://docs.microsoft.com/en-us/dotnet/api/system.reflection.emit.ilgenerator.beginexceptfilterblock?view=netframework-4.8

stonewell:
I found the document too, but from the stackTrace, the exception is thrown when EndException called not BeginExceptionFilterBlock, that's why I trying to fix the problem instead find other solution. my fix is based on this document [https://docs.microsoft.com/en-us/previous-versions/dotnet/netframework-3.0/h0x241a0(v=vs.85)]

in the document, MS said
Emit some MSIL.
BeginExceptionBlock
Emit the MSIL for the try block.
BeginExceptFilterBlock
Emit the MSIL for the filtered exception handler.
BeginCatchBlock
Emit the MSIL for the catch block. The catch handler should be supplied with a null type.
BeginFaultBlock
Emit the MSIL for the fault block.
BeginFinallyBlock
Emit the MSIL for the finally block.
EndExceptionBlock

pardeike:
My unit tests test for multiple .NET versions and unfortunately, at least one of them fails in BeginExceptionFilterBlock. Harmony is used with many different .NET versions so a solution that only works for specific ones isn't useful.

mitchcapper:
I assume it isn't worth trying to detect this and throw an exception that is clearer, but if anyone else gets the `System.InvalidOperationException : Incorrect code generation for exception block` error check the code you probably will have a try/ catch with a when clause on it in the original code.

PR #188, opened 2019-05-13
improve wording on exception for pass through postfix

Mehni:
The exception came up a few times on the RimWorld discord. This should clear up some confusion.


pardeike:
Sure!

PR #206, opened 2019-09-01
add a convenient factory to generate transpilers

waingt:
I find quite of my code is duplicate when using harmony transpiler to patch.
So i made a TranspilerFactory because i didn't find something similar to this in Tools.
It can parse input string with wildcards to generate transpilers to replace,insert and delete.
Here is the usage:
```csharp
class Student
{
    public Student(string name)
    {
        Name = name;
    }
    private string Name;
    public void Speak(string something)
    {
        Console.WriteLine($"{Name} says : {something}");
    }
}
[HarmonyPatch(typeof(Student), "Speak")]
static class StudentSpeakPatch
{
    static IEnumerable<CodeInstruction> Transpiler(ILGenerator generator, IEnumerable<CodeInstruction> instr)
    => new TranspilerFactory().Delete("ldstr  says : ").Insert("ldstr  said nothing")
    .Replace("ldarg.1;call *", "call string::Concat(string, string)").Transpiler(generator, instr);
}
```
Part of my idea comes from /pull/9

### DIFF
--- a/Harmony/Public/PatchProcessor.cs
+++ b/Harmony/Public/PatchProcessor.cs
@@ -265,9 +265,9 @@ namespace HarmonyLib
 
 		void PrepareType()
 		{
-			var mainPrepareResult = RunMethod<HarmonyPrepare, bool>(true);
-			if (mainPrepareResult == false)
-				return;
+			//var mainPrepareResult = RunMethod<HarmonyPrepare, bool>(true);
+			//if (mainPrepareResult == false)
+			//	return;
 
 			var originalMethodType = containerAttributes.methodType;
 
--- a/Harmony/Tools/AccessTools.cs
+++ b/Harmony/Tools/AccessTools.cs
@@ -12,26 +12,31 @@ namespace HarmonyLib
 	/// <summary>A helper class for reflection related functions</summary>
 	public static class AccessTools
 	{
-		/// <summary>Shortcut to simplify the use of reflections and make it work for any access level</summary>
-		public static BindingFlags all = BindingFlags.Public
-			| BindingFlags.NonPublic
-			| BindingFlags.Instance
-			| BindingFlags.Static
-			| BindingFlags.GetField
-			| BindingFlags.SetField
-			| BindingFlags.GetProperty
-			| BindingFlags.SetProperty;
+        /// <summary>Shortcut to simplify the use of reflections and make it work for any access level</summary>
+        public static BindingFlags all = BindingFlags.Public
+            | BindingFlags.NonPublic
+            | BindingFlags.Instance
+            | BindingFlags.Static
+            | BindingFlags.GetField
+            | BindingFlags.SetField
+            | BindingFlags.GetProperty
+            | BindingFlags.SetProperty
+            | BindingFlags.FlattenHierarchy;
 
 		/// <summary>Shortcut to simplify the use of reflections and make it work for any access level but only within the current type</summary>
 		public static BindingFlags allDeclared = all | BindingFlags.DeclaredOnly;
-
-		/// <summary>Gets a type by name. Prefers a full name with namespace but falls back to the first type matching the name otherwise</summary>
-		/// <param name="name">The name</param>
-		/// <returns>A Type</returns>
-		///
-		public static Type TypeByName(string name)
+        static readonly Dictionary<string, Type> KeywordTypes = new Dictionary<string, Type>{ { "bool", typeof(bool) }, { "byte", typeof(byte) }, { "char", typeof(char) },{"decimal",typeof(decimal)
+},{"double",typeof(double)},{"float",typeof(float)},{"int",typeof(int)},{"long",typeof(long)},{"sbyte",typeof(sbyte)},{"short",typeof(short)},{"string",typeof(string)},{"uint",typeof(uint)},{"ulong",typeof(ulong)},{"ushort",typeof(ushort)}};
+        /// <summary>Gets a type by name. Prefers a full name with namespace but falls back to the first type matching the name otherwise</summary>
+        /// <param name="name">The name</param>
+        /// <returns>A Type</returns>
+        ///
+        public static Type TypeByName(string name)
 		{
-			var type = Type.GetType(name, false);
+            Type type;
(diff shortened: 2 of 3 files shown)


pardeike:
That’s a lot of code and no clear description on how to use it. Your example is not sufficient because your API is string based and thus need a full specification of the specific domain language you used/invented. 

As it is right now I cannot incorporate your code.

waingt:
Now it is definitely more clear and convenient! 
Cant wait to be added😝

waingt:
Usage of this version:
- targetClass:
```cs
class Student
{
    public Student(string name) => Name = name;
    private string Name;
    public void Speak(string something) => Console.WriteLine($"{Name} says : {something}");
}
```
- patch1:
```cs
var harmony = new Harmony("zzz");
new TranspilerFactory()
    .Search("ldfld Student::Name")
    .Insert("localvar string;stloc.s 0;br 0")
    .Replace("ldarg.1", "label 0;ldarg.1;ldstr  is said by ;ldloc.s 0")
    .PatchFor(harmony, "Student::Speak(string)");
```
- patch2:
```cs
[HarmonyPatch(typeof(Student), "Speak")]
class MyPatch : TranspilerPatch
{
    static void Prepare() => factory.Search("ldfld Student::Name")
        .Insert("localvar string;stloc.s 0;br 0")
        .Replace("ldarg.1", "label 0;ldarg.1;ldstr  is said by ;ldloc.s 0");
}
```
Both of them works!

pardeike:
Maybe you misunderstood me, so let me rephrase it. I need full documentation on the methods and the arguments of all methods. As it is now, it is not clear to me what I can pass to your methods - especially the string arguments.

Some examples (but there are many more questions):

- what happens if I call `.Search()` twice in a row?
- what does `PatchFor` do?
- are there more methods than the ones you listed in the example?
- can I pass `”Andreas”` in any argument and if not, what will happen?
- what are the search arguments?
- how do they work?
- how do the methods work together?
- what are the limits?

If your pull request would contain a simple self-explaining method or just a bug fix, I could pull it right away. Here, you give me a whole new way to run stuff but no explanation. I could put a lot of time reading through your code and figuring it out myself but just expecting me to accept this or understand your design from just the code is asking too much.

waingt:
Sorry I do have a bad habit that write very few comments. A full documentation is little bit hard to me.
But we programmers communicate with code, aren't we😅
Some explaintion:
- My main idea is to treat the CodeInstuctions more as a stream instead of an array. It is only processed once from beginning to end.
- Once you add a Search task, the position of the stream is moved forward. This is the only thing that it does.
- So you can't Search backwards, the position never goes back. It will search the rest of the stream
- PatchFor just calls Harmony.Patch(), the generated transpiler will be applied to specific method
- If you pass ”Andreas” in Parse.Parse(), it trys to get a coresponding InlineNone OpCode and, of course, throw a Exception
- I tried to make it parse like logged ILCode's style

pardeike:
But as a minimum you **have to** explain the string syntax. Completely, on all calls and for all cases. I understand that you find it difficult to document this but I won’t pull it otherwise.

waingt:
It's hard to explained in the comments. But i can do here
- Format for single instruction
`"opcode operand(optional)"`
both opcode and operand can be * , meaning arbitrary
- Format for instructions
`"[instructions seperated by semicolon]“`

opcode type | format for operand
-------------- | -------------------
InlineMethod(call,newobj) | [type name][one or two colon][method name]\(optional)([parameter types seperated by comma])(optional)
InlineField(ldfld,stfld) | [type name][one or two colon][field name]
DeclareVar(localvar) | string for localvar type
InlineVar(ldloc,stloc_s) | number for localvar index
MarkLabel(label) | number for label index
InlineBrTarget(br,brtrue) | number for label index
InlineString(ldstr) | just the string, no need to add quotation marks
InlineI etc. | will be converted to certain value type

**Examples**
- call string::Concat(string, string)
- newobj Student(string)
- br 0
- label 0
- localvar int
- ldloc_s 0

waingt:
Just want to shared some code with you
```cs
new TranspilerFactory()
.Delete("ldloc_s *;call ForbidUtility::SetForbiddenIfOutsideHomeArea(Thing)")
.PatchFor(harmony, "Pawn::Kill")
.ToString(s => Debug(s))
.ClearAllTasks()
.Replace("ldc.i4.1;callvirt Pawn_EquipmentTracker::DropAllEquipment(IntVec3, bool);", "ldc.i4.0;callvirt Pawn_EquipmentTracker::DropAllEquipment(IntVec3, bool);")
.Replace("ldc.i4.1;ldc.i4.0", "ldc.i4.0;ldc.i4.0")
.PatchFor(harmony, "Pawn::DropAndForbidEverything")
.ToString(s => Debug(s))
;
```

waingt:
Just want to shared some code with you
```cs
new TranspilerFactory()
.Delete("ldloc_s *;call ForbidUtility::SetForbiddenIfOutsideHomeArea(Thing)")
.PatchFor(harmony, "Pawn::Kill")
.ToString(s => Debug(s))
.ClearAllTasks()
.Replace("ldc.i4.1;callvirt Pawn_EquipmentTracker::DropAllEquipment(IntVec3, bool);", "ldc.i4.0;callvirt Pawn_EquipmentTracker::DropAllEquipment(IntVec3, bool);")
.Replace("ldc.i4.1;ldc.i4.0", "ldc.i4.0;ldc.i4.0")
.PatchFor(harmony, "Pawn::DropAndForbidEverything")
.ToString(s => Debug(s))
;
```

lbmaian:
My 2 cents: I don't think this belongs in Harmony proper - it should be its own library.

It's a rather opinionated take on transpiler abstraction. I appreciate the overall functionality, but I'm not a fan of the string-based API. I'd much prefer a more type-safe API (for ex, something like the [internal ILGen API](https://referencesource.microsoft.com/#system.core/Microsoft/Scripting/Compiler/ILGen.cs)). And some functionality like the searching could be generified to work on any list rather than a list of CodeInstructions.

pardeike:
I totally agree with @lbmaian and will reject this pull.

PR #207, opened 2019-09-14
Support target methods without body

mcpiroman:
This allows to patch invocations to `extern` P/Invoke methods, e.g. if a game uses native plugin.

### DIFF
--- a/Harmony/Internal/CodeTranspiler.cs
+++ b/Harmony/Internal/CodeTranspiler.cs
@@ -195,8 +195,13 @@ namespace HarmonyLib
 		{
 			var type = transpiler.GetParameters()
 				  .Select(p => p.ParameterType)
-				  .FirstOrDefault(t => t.IsGenericType && t.GetGenericTypeDefinition().Name.StartsWith("IEnumerable", StringComparison.Ordinal));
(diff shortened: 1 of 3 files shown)


pardeike:
So this basically returns an empty list of instructions for an external method? Sounds better than just throwing. I still wonder if one would want to fake a default return value and a Ret. Not sure, maybe a clean empty list is better.

mcpiroman:
I guess empty list is better in that faking things is more likely to cause trouble and confuse people.

So since in such cases original instructions list is empty, I've made this transpiler's argument optional (which can also be useful when original body is not required).

What I didn't mention is it doesn't support prefixes or postfixes (as there is no 'original' function to call). This would be theoretically possible but would require to copy the signature of a original method, which to my knowledge  is quite hard to do (parameter attributes in particular).

LunarWhisper:
Is there an approximate implementation timeline?

pardeike:
I still need to quality check this pull request and find out why it makes the tests fail. So for now there is no estimate

pardeike:
This PR touches a very sensitive part of Harmony and probably breaks parallel compatibility. It needs a lot more work and verification to get accepted.

PR #219, opened 2019-11-10
Allow transpiler method parameter for the instructions IEnumerable be a ICollection, IList, or List

lbmaian:
I find myself almost always converting the IEnumerable<CodeInstruction> into a List<CodeInstruction> for access to the List methods like FindIndex, and from what I can tell of the harmony code, it's internally such a List anyway.

I generalized it to all the generic interfaces the List<T> implements.

Aside: I'm not sure how you get unit tests running yourself, but in order to run them via Visual Studio Test Explorer on all target frameworks (except for netcoreapp3.0), I had to locally remove the Microsoft.NET.Test.Sdk and NUnit3TestAdapter nugets, and install the NUnit 3 Test Adapter extension. Microsoft.NET.Test.Sdk requires netcoreapp2.1+ or net45+, while NUnit3TestAdapter requires netcoreapp1.0+, so they don't work for the .NET Framework versions like net35.

### DIFF
--- a/Harmony/Internal/CodeTranspiler.cs
+++ b/Harmony/Internal/CodeTranspiler.cs
@@ -194,8 +194,10 @@ namespace HarmonyLib
 		internal static IEnumerable ConvertToGeneralInstructions(MethodInfo transpiler, IEnumerable enumerable, out Dictionary<object, Dictionary<string, object>> unassignedValues)
 		{
 			var type = transpiler.GetParameters()
-				  .Select(p => p.ParameterType)
-				  .FirstOrDefault(t => t.IsGenericType && t.GetGenericTypeDefinition().Name.StartsWith("IEnumerable", StringComparison.Ordinal));
+				.Select(p => p.ParameterType)
+				.FirstOrDefault(t => t.IsGenericType && t.GetGenericTypeDefinition().Name is var name &&
+					(name.StartsWith("IEnumerable", StringComparison.Ordinal) || name.StartsWith("ICollection", StringComparison.Ordinal) ||
+					name.StartsWith("IList", StringComparison.Ordinal) || name.StartsWith("List", StringComparison.Ordinal)));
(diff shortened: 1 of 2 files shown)


pardeike:
Is this backwards compatible? I guess not.

lbmaian:
> 
> 
> Is this backwards compatible? I guess not.

It is. See the unit tests.

It looks like the build is already failing regardless of my change, perhaps because of the reasons in my "aside".

lbmaian:
Yeah... looking at the [commit history][1], I'm seeing a string of red 'X's indicating build failures, and looking at them, for the same reason the above checks failed.

[1]: https://github.com/pardeike/Harmony/commits/master

pardeike:
You missed my point. I was talking about multiple versions of Harmony patching the same method. And in any order. Since transpilers are chained and will be called across assemblies you can easily end up with Harmony 1 calling Harmony 2’s transpilers...

lbmaian:
I don't see how my patch would affect that at all. It just allows transpiler method definitions to use ICollection/IList/List instead of IEnumerable.

If you're referring to the UnpatchAll I added at the end of the unit test, that was merely to reuse the same original method and transpiler testing across all transpiler method definition variants. I could have instead created separate original methods to patch for each transpiler method definition variant, but that's tedious and it wasn't clear to me how you want your text fixtures (Class<n> with method <x>) laid out. This patch has nothing to do with patching the same method with multiple transpilers - that functionality is unchanged.

lbmaian:
Discussed in discord. To clarify, the concern is not with patching the same method in the unit test. It's about Harmony 1 having to interact with Harmony 2 patches for the same original method, and Harmony 1 failing on transpilers of an API it's not aware of (as when patching the same original method, it essentially has to replay all patches on that method).

So it's a backwards compatibility issue that's a problem for Harmony 2 API-breaking changes in general.

pardeike:
Sorry, I decided to keep the IEnumerable<> interface. It's not hard to work around.

PR #220, opened 2019-11-17
Fix rare MethodInvoker directBoxValueAccess=false bug where an inopportune GC can cause reboxed value address to become invalid

lbmaian:
## Background
Some background, continuing from the discussion on discord:

I was curious as to how and why the `directBoxValueAccess=false` logic in `MethodInvoker` works for by-ref value types. I learned the whole point of this mode is to avoid directly updating the boxed value in the arguments array, which could which be reflected in all existing references to that boxed value (not just in the arguments array), and to instead replace that boxed value in the arguments array with a "reboxed" value, such that updates to the boxed value are instead localized only to this reboxed value.

To illustrate the difference:
```cs
public static void TimesTwo(ref int x) => x *= 2;
...
var method = GetType().GetMethod(nameof(TimesTwo));
var handler = MethodInvoker.Handler(method, method.DeclaringType.Module, directBoxValueAccess);

var boxedValue = (object)1000;
var args = new object[] { boxedValue };
handler(null, args);
Assert.AreEqual(2000, (int)args[0]);
Assert.AreEqual(directBoxValueAccess ? 2000 : 1000, (int)boxedValue);
```

One of the techniques used in `directBoxValueAccess=false` mode is to store the address of the unboxed value into a "pinned" unmanaged `void*` pointer. The point of this is to avoid need to create local variables for each by-ref value type, in favor of a single local. It's pinned to ensure that unboxed value that it points to doesn't change its address (via garbage collections), which would invalidate an unmanaged pointer.

## Problem

This doesn't work however. Unmanaged pointer locals apparently can't be directly pinned. Rather, there needs to a managed reference local that's pinned, and the unmanaged pointer needs have an address at/within the data this managed reference refers to.

For example, the following:
```cs
struct S { ... }
...
void Foo()
{
	var arr = new S[] { new S() };
	unsafe
	{
		fixed (void* ptr = &arr[0]) { ... }
	}
}
```
generates the following locals for `Foo`:
```
.locals init (
	[0] valuetype S[],
	[1] valuetype S& pinned,
	[2] void*
)
```
Notice how `arr[0]` is effectively assigned to a pinned local of type `S&`, and that the local of type `void*` isn't pinned at all.

Another note: Managed references (like `<type>&` and object locals) and managed pointers (like the result of the `unbox` instruction) have internal pointers to the actual memory addresses that are automatically updated if the data at those memory addresses are moved (via garbage collections). The notion of pinning is only required within "unmanaged" code (which includes the usage of unmanaged pointers like this `void*` local).

In practice, it's likely extremely rare, and perhaps impossible on some .NET runtimes, for the memory address of the reboxed value to change between its creation (during the setting up of the method invocation arguments) and the method invocation, but I've confirmed with tests that it's technically possible.

## Solution

A potential solution is to add an additional pinned local for `<valuetype>&`, i.e. managed reference for the value type, but since it's type-specific, it would result in multiple locals, defeating the purpose of the single `void*` local.

Rather than pinned locals for `<valuetype>&`, how about a generic pinned local for the boxed value? The boxed value is always an object, so we only need a single pinned local of type `object` for this to work.

That works. But if we now have a single local that stores the boxed object, we actually no longer need that single `void*` local. Just store the reboxed object in this boxed object local, update the argument array with the reboxed object as usual, load the reboxed object, then `unbox` it.

Furthermore, if we get rid of the `void*` local, we no longer have an unmanaged pointer, and thus the local for the boxed object doesn't need to be pinned at all.

Of course, if the method being invoked is itself unmanaged code, then yes, the local would have to be pinned. But that applies to *all* pointer arguments to such a method - all pointer arguments, regardless of whether it's a value type or not, would need to be stored in pinned locals. Since `MethodInvoker` isn't doing that, the implicit assumption is that the method being invoked is either not unmanaged code, or doesn't access any managed references that `MethodInvoker` could be aware of.

## Code change notes

To discover the above problem, I had to put together a test that shows it. I did this by make `MethodInvoker` subclassable, and having a test subclass of `MethodInvoker` (named `MethodInvokerGCHell`) that interweaves in instruction `call`s to a `TryMoveAddressesViaGC` method that triggers GCs in a way that should compact the heap and thus move data to different memory addresses:
```cs
		static void TryMoveAddressesViaGC()
		{
			var memoryPressureBytesAllocated = 100000000;
			GC.AddMemoryPressure(memoryPressureBytesAllocated);
			GC.Collect();
			GC.RemoveMemoryPressure(memoryPressureBytesAllocated);
		}
```

The tests are run under the following scenarios:
1. `directBoxValueAccess=false`, use normal `MethodInvoker`
2. `directBoxValueAccess=true`, use normal `MethodInvoker`
3. `directBoxValueAccess=false`, use test `MethodInvokerGCHell`
4. `directBoxValueAccess=true`, use test `MethodInvokerGCHell` (fails before the fix)

I also added a bunch of `#if TRACE`-guarded debugging code. As they do clutter up the code, I can remove them upon request.

### DIFF
--- a/Harmony/Extras/MethodInvoker.cs
+++ b/Harmony/Extras/MethodInvoker.cs
@@ -4,7 +4,7 @@ using System.Reflection.Emit;
 
 namespace HarmonyLib
 {
-	// Based on https://www.codeproject.com/Articles/14593/A-General-Fast-Method-Invoker
+	// Based on https://github.com/MonoMod/MonoMod/blob/master/MonoMod.Utils/FastReflectionHelper.cs
 
 	/// <summary>A delegate to invoke a method</summary>
 	/// <param name="target">The instance</param>
@@ -23,7 +23,7 @@ namespace HarmonyLib
 		///
 		public static FastInvokeHandler GetHandler(DynamicMethod methodInfo, Module module)
 		{
-			return Handler(methodInfo, module);
+			return GetHandler(methodInfo, module);
 		}
 
 		/// <summary>Creates a fast invocation handler from a method and a module</summary>
@@ -32,10 +32,10 @@ namespace HarmonyLib
 		///
 		public static FastInvokeHandler GetHandler(MethodInfo methodInfo)
 		{
-			return Handler(methodInfo, methodInfo.DeclaringType.Module);
+			return GetHandler(methodInfo, methodInfo.DeclaringType.Module);
 		}
 
-		static FastInvokeHandler Handler(MethodInfo methodInfo, Module module, bool directBoxValueAccess = false)
+		public static FastInvokeHandler GetHandler(MethodInfo methodInfo, Module module, bool directBoxValueAccess = false)
 		{
 			var dynamicMethod = new DynamicMethod("FastInvoke_" + methodInfo.Name + "_" + (directBoxValueAccess ? "direct" : "indirect"), typeof(object), new Type[] { typeof(object), typeof(object[]) }, module, true);
 			var il = dynamicMethod.GetILGenerator();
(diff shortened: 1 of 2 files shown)


lbmaian:
Rebased onto master (now that the CI actually works).

Went ahead and deleted the `#if TRACE`-guarded debugging code.

lbmaian:
Well, that's interesting. An unrelated unit test (AttributeReversePatches.TestReversePatchingWithAttributes) spuriously caused a crash in one of the CI mono builds (https://travis-ci.com/pardeike/Harmony/jobs/258767361#L632) but succeeded in the next (https://travis-ci.com/pardeike/Harmony/builds/137469613). Might indicate that reverse patching is unreliable in mono?

Also added XML documents to fix warnings. This is ready for review now.

PR #221, opened 2019-11-18
Update nugets, try updating CI to run tests for all target frameworks

lbmaian:
1. Update nugets to latest versions
2. Add Microsoft.NETFramework.ReferenceAssemblies nuget dependency for non-Windows builds
3. Update appveyor config:
    1. Use 'Visual Studio 2019' image
    2. Run dotnet test for all target frameworks except .NET Framework 3.5
    3. Run nunit3-console for .NET Framework 3.5 with --inprocess flag, because Microsoft.NETFramework.ReferenceAssemblies doesn't include a build for .NET Framework 3.5 yet
4. Update travis config:
    1. Use dotnet 3.0 and latest nunit3-console nuget
    2. Run dotnet test for all target frameworks except .NET Framework 3.5
    3. Run nunit3-console for .NET Framework 3.5 with --inprocess flag, because Microsoft.NETFramework.ReferenceAssemblies doesn't include a build for .NET Framework 3.5 yet
5. Update azure-pipelines to use 'windows-latest' image, but although it now doesn't fail, it still does practically nothing as far as I can tell

PR #222, opened 2019-11-20
Fix remaining unit testing issues (excluding netcoreapp3.0 issues)

lbmaian:
* Fix DynamicArgumentPatches.SendingArguments unit test on mono (see #221 for the error)
* Don't enable Harmony.DEBUG in unit tests, allow linux CI to fail if it is enabled


pardeike:
Awesome

PR #223, opened 2019-11-25
Porting Harmony to .NET Core 3

berkeleyinc:
Hey, 

I've been working on porting Harmony to .NET Core 3.
The main issue was that MethodInfo objects are not longer serializable in .NET Core

To bypass that, i've created a MethodInfoWrapper that traverses the RuntimeMethodInfo object and performs a shallow copy.
gg

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,28 +1,79 @@
-dist: xenial
-addons:
-  snaps:
-  - name: dotnet-sdk
-    classic: true
-    channel: latest/beta
-sudo: required
+# Note: Even though dotnet test works in .NET Framework 4+, it apparently still requires mono.
+# dotnet test currently doesn't work for .NET Framework 3.5, because Microsoft.NETFramework.ReferenceAssemblies
+# doesn't have a build for it yet: https://github.com/dotnet/core-sdk/issues/2022
+# Also, builds that use mono are currently very slow due to https://github.com/travis-ci/travis-ci/issues/4571
+
 language: csharp
+solution: Harmony.sln
+
 os:
   - linux
   - osx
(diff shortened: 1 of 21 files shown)


lbmaian:
Quick comment, not actually reviewing the code - just noticed the CI failed (when it's just been recently fixed) and looked into it:

You can't just remove the `net35`, `net45`, `net472`, `net48`, and (maybe) `netstandard2.1` target frameworks. Especially not when major users of Harmony include mods for games that still require the `net35` build like RimWorld.

If you have new code that requires `netcoreapp3.0`, delimit that code in a `#if NETCOREAPP3_0`/(`#else`)/`#endif` (replace `NETCOREAPP3_0` with `NETCOREAPP` if it works on all .NET Core versions).

mitchcapper:
@berkeleyinc  thanks for this, works fantastic.

lbmaian:
The CI failures indicate that the changes aren't compatible with .NET Framework 3.5, which doesn't support `dynamic`, nor has `System.Diagnostics.Contracts`.

And yes, there needs to be a package reference to Microsoft.CSharp but only for .NET Core builds.

So there's still cross-target-framework work to be done.

.NET Core specific code needs to be guarded with `#if NETCOREAPP`/`#endif`.

New package references (and any other additions to the .csproj) need to be guarded with `Condition="$(TargetFramework.StartsWith('netcoreapp'))"` attribute (or if it's really specific .NET Core 3.0, `Condition="'$(TargetFramework)' == 'netcoreapp3.0'"`).

pardeike:
My latest changes to master make this PR obsolete. I also made sure that it's backwards compatible and that all tests success.

Issue #215, opened 2019-11-01
Traverse.Property does not find properties from base types


ManlyMarco:
**Describe the bug**
Traverse.Property does not find properties from base types of the object instance passed in to Traverse.Create.

**To Reproduce**
```cs
class A : B { }
class B { public bool Test {get;} = true; }

Traverse tr = Traverse.Create(new A());
var prop = tr.Property("Test");
prop.GetValue<bool>() // Fails silently
```

**Expected behavior**
Traverse.Property should find properties from base classes. When passing in an instance it's not possible (or desirable) to specify the correct type to get the property.

**Runtime environment (please complete the following information):**
 - Windows 10, 64bit
 - .NET 4.5 in Unity 2018
 - Harmony version 2.0.0.0

pardeike:
I am willing to make Traverse look for inherited fields, properties and methods but I am a bit sceptical about backwards compatibility. I think within the frame of the 2.0 release it's ok.

pardeike:
Fixed in https://github.com/pardeike/Harmony/commit/6b7b691b8c2d6047e59c466a4dc0ff3a251d3b0a

Issue #213, opened 2019-10-20
Traverse.Property doesn't work for static properties


ManlyMarco:
**Describe the bug**
Traverse.Property doesn't work for static properties when using the Type overload of Traverse.Create instead of the instance overload.

**To Reproduce**
1. Have a class with a static property.
2. Use Traverse.Create(typeof(MyClass)).Property to get the static property.
3. Property check fails at https://github.com/pardeike/Harmony/blob/master/Harmony/Tools/Traverse.cs#L272 while the field check works fine.

**Expected behavior**
Traverse.Property should work for static types like Traverse.Field does.

pardeike:
Fixed in master

PR #226, opened 2019-11-28
Make Traverse.Property, Traverse.Field and Traverse.Method behavior uniform

ManlyMarco:
Made traverse detect static and non-declared properties, as well as non-declared fields (static fields were already detected). This makes `Traverse.Field` and `Traverse.Property` work the same way as `Traverse.Method`.

Fixes issues #215 and #213.


pardeike:
Cannot accept this because changing the AccessCache to look in base classes would break all kind of things. I will close this but work on an alternative solution to archive the same result.

PR #228, opened 2019-12-12
Fix MethodBodyReader throwing error when parsing local var operand and no generator

lbmaian:
Fix MethodBodyReader throwing error when parsing local var operand and no generator.

Add unit test.

In the future, there should be a way to publicly expose this functionality (something like `List<CodeInstruction> GetMethodInstructions(MethodBase method)`) along with caching.

### DIFF
--- a/Harmony/Internal/MethodCopier.cs
+++ b/Harmony/Internal/MethodCopier.cs
@@ -561,7 +561,7 @@ namespace HarmonyLib
 							else
 							{
 								instruction.operand = lvi;
-								instruction.argument = variables[lvi.LocalIndex];
+								instruction.argument = variables?[lvi.LocalIndex] ?? lvi;
 							}
 						}
 						else
@@ -583,7 +583,7 @@ namespace HarmonyLib
 							else
 							{
 								instruction.operand = lvi;
-								instruction.argument = variables[lvi.LocalIndex];
+								instruction.argument = variables?[lvi.LocalIndex] ?? lvi;
 							}
 						}
 						else
--- a/HarmonyTests/IL/TestMethodBodyReader.cs
+++ b/HarmonyTests/IL/TestMethodBodyReader.cs
@@ -1,0 +1,68 @@
+using System.Collections;
+using System.Reflection;
+using System.Reflection.Emit;
+using HarmonyLib;
+using HarmonyLibTests.Assets;
+using NUnit.Framework;
+
+namespace HarmonyLibTests.IL
+{
+	[TestFixture]
+	public class TestMethodBodyReader
+	{
+		[Test]
+		public void CanGetInstructionsWithNoILGenerator()
+		{
+			var method = typeof(Class12).GetMethod(nameof(Class12.FizzBuzz));
+			var instrsNoGen = MethodBodyReader.GetInstructions(generator: null, method);
+
+			var dynamicMethod = DynamicTools.CreateDynamicMethod(method, "_Patch");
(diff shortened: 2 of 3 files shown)


pardeike:
I have a feeling that this would easily lead to lazy people use index values which is against common interest to keep things compatible and robust. The RimWorld community has proven over and over again that poorly written mods usually make life hell for everyone else.

lbmaian:
I don't particularly care about having to publicly expose instructions outside of transpilers, but my current use case and solution relies on this at the moment.

My own use case for this is for identifying the proper compiler-generated methods without have to hardcode in the hidden type+method names determined from a decompiler. I find such hardcoding unnecessarily brittle, since a recompile of the assembly being runtime-patched could easily change the hidden names.

Ideally, a target method would be somehow partially decompiled such that compiler-generated types+methods look "inlined", but that's obviously really difficult to do. (And while decompilers may show the C# code, lambdas and all, all in a single method, they typically don't do the same when showing IL.)

Another alternative would be to have a transpiler that can identify the compiler-generated type+method that itself needs to be targeted. But in Harmony's declarative form (attributes + `PatchAll`), there's no way to sequentially patch like this - `HarmonyTargetMethod(s)` of one patch class cannot depend on the transpiler of another patch class being run. Worse yet, if that initial transpiler is only used for finding such a type+method and doesn't modify the method, that's a completely wasteful patch being applied to that method being transpiled.

So without sequential patching, my current solution is some helper extension methods that can filter for compiler-generated methods (https://gist.github.com/lbmaian/9daf96b4aa4db1970d5b608f0401df69) that accept an additional filter. I use it like this:

```cs
		[HarmonyTargetMethods]
		static IEnumerable<MethodBase> CalculateMethods(HarmonyInstance harmony) =>
			typeof(ChoiceLetter_RansomDemand).FindLambdaMethods(method =>
				method.ReturnType == typeof(void) &&
				method.GetParameters().Length == 0 &&
				// Note: RimWorld was compiled in a way such that the lambda methods are called "<>m__x" where x is a number.
				// In case RimWorld is ever recompiled in a different way, only going to assume that lambda methods start with "<".
				method.Name.StartsWith("<") &&
				HasSilverInTradeBeaconRangeMethod(method))
			.Prepend(typeof(ChoiceLetter_RansomDemand).FindIteratorMethod(enumeratorType =>
				typeof(IEnumerable<DiaOption>).IsAssignableFrom(enumeratorType)));

		static bool HasSilverInTradeBeaconRangeMethod(MethodInfo method)
		{
			var instructions = MethodReader.GetInstructions(method);
			return instructions.Any(instruction => instruction.opcode == OpCodes.Call && IsTradeBeaconMethod(instruction.operand as MethodInfo));
		}
```
(`MethodReader` is a wrapper over `MethodBodyReader` that allows no `ILGenerator`.)

I'm not attached to this solution, since you're right, naive usage of `MethodBodyReader` could be harmful, but if it's this or hardcoding, I'd rather choose the former.

To conclude, for a target method, there should be a way to get compiler-generated methods for lambdas and iterators without hardcoding hidden names and without needing a wasteful Harmony transpiler patch, and use them as Harmony method targets. I'm currently relying on reading the method's instructions via `MethodBodyReader` to do this.

lbmaian:
Also, this PR itself doesn't expose anything public that wasn't public before. It merely allows `MethodBodyReader` to be used internally without an `ILGenerator`.

pardeike:
Alright, I agree.

PR #229, opened 2019-12-13
dotnet build now works cross-platform (but not dotnet test), misc

lbmaian:
* Add workaround .NET Framework 3.5 reference assemblies (fixes x-platform dotnet build)
* Update to non-preview Microsoft.NETFramework.ReferenceAssemblies
* Update comment on why dotnet test doesn't work for .NET Framework 3.5
* Fix copy&paste error in MethodBodyReader unit test


lbmaian:
It looks like Travis CI is broken for Linux for .NET Core 3.0 users, and it's not just us: https://travis-ci.community/t/unable-to-locate-package-dotnet-sdk-3-0-on-linux/6443

It's pretty clear by now that Travis is pretty shit at .NET/C# CI. Think we should migrate off Travis to Appveyor, considering that the latter apparently supports Linux and OSX now.

PR #245, opened 2020-02-02
Add functional basic inline signature parsing and emitting

0x0ade:
This PR adds support for parsing `InlineSig`-typed operands as `InlineSignature`s, and re-emitting them using `MonoMod.Common`.

Right now parsing works perfectly fine and emitting using Cecil to dump the round-tripped IL returns almost exactly the same IL as `MonoMod.Common`, except for a few `.s`hort form OpCode differences.

Sadly, the emitted method doesn't want to run at all right now, and I don't understand why. To make matters worse: Only the `DynamicMethod` backend supports emitting the `calli` that I'm testing this with in the first place.

If you want to try it out yourself: Find a moddable XNA game and run [this Harmony patch](https://gist.github.com/0x0ade/840b3fcfecc5f7627e654353cc320460) in it.

This PR also bumps the MonoMod.Common dependency to `20.2.1.1`.


pardeike:
I bumped the version already (and added a comment to the azure repo into csproj

pardeike:
May want to change the version number to 20.2.2.1 so it does not conflict

0x0ade:
Merged the latest changes from master, it now no longer throws absurd errors! Seems like they were caused by the memory corruption fixed by #247.

PR #247, opened 2020-02-03
Fix variable definition order: keep orig vars first

0x0ade:
Commit f213218d30237cee09085ffe60cf6d0a94501b71 introduced a bug in which a method's original variables got defined after the temporary result variable and all other patch-related variables, resulting in (best case) the JITter throwing an `ExecutionEngineException` and (worst case) memory corruption in form of an uncatchable access violation.

This PR fixes the variable definition order.


pardeike:
Well done!

PR #257, opened 2020-02-24
Fix CodeInstruction Extension IsLdarg for short-form operands

alextd:
Ldarg_s needs the operand cast to Byte, not int - InvalidCastException otherwise.

(There are probably more occurrences of this problem, but this is just the one I hit)

Example transpiler simply to catch the exception:
```cs

[HarmonyPatch(typeof(PlaceWorker_PreventInteractionSpotOverlap), nameof(PlaceWorker_PreventInteractionSpotOverlap.AllowsPlacing))]
public static class IgnoreInteractionSpotOverlap {
public static IEnumerable<CodeInstruction> Transpiler(IEnumerable<CodeInstruction> instructions)
{
	foreach(var inst in instructions)
	{ 
		try
		{
			bool test = inst.IsLdarg(5);
		}
		catch (Exception e)
		{
			Log.Message($"- {inst}/{inst.operand.GetType()}");
			Log.Message($"- {e}");
		}
		yield return inst;
	}
}
}
```

results:
- ldarg.s 4/System.Byte
- `System.InvalidCastException: Specified cast is not valid.
  at HarmonyLib.CodeInstructionExtensions.IsLdarg (HarmonyLib.CodeInstruction code, System.Nullable`1[T] n) [0x000d1]`

### DIFF
--- a/Harmony/Tools/Extensions.cs
+++ b/Harmony/Tools/Extensions.cs
@@ -172,8 +172,9 @@ namespace HarmonyLib
 			if ((n.HasValue == false || n.Value == 1) && code.opcode == OpCodes.Ldarg_1) return true;
 			if ((n.HasValue == false || n.Value == 2) && code.opcode == OpCodes.Ldarg_2) return true;
 			if ((n.HasValue == false || n.Value == 3) && code.opcode == OpCodes.Ldarg_3) return true;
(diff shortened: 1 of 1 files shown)


pardeike:
Thanks, checking for other occurrences

PR #261, opened 2020-02-27
Remove unnecessary DefineConstants in csprojs again

lbmaian:
Remove unnecessary DefineConstants in csprojs again


pardeike:
sorry

PR #280, opened 2020-04-04
Fix problems with methods that return structs

### LINKED ISSUE #277: Problem with patching virtual method that returns UnityEngine.Color
**Describe the bug**
I got various kinds of errors while trying to patch a certain method with 2.0.0.9.

The same patch works with 1.2.0.1.

~~One very suspicious thing is that in the harmony.log IL code of the replacement method, there seem to be 27 local vars defined (index 0 to 26).~~

~~In the harmony-generated IL after `// end original`, there are statements like hat reference local var with index 27 (doesn't exist?????):~~

Edit: I checked again and it seems var with index 27 is just not listed because the list is generated while the original method IL is printed. So I'm pretty clueless what is happening...

I don't really get why you need var 27 when you already have var 25 which stores exactly the same value

```
IL_07C1: Label0
IL_07C1: ldloc      25 (UnityEngine.Color)
IL_07C5: ldloca     27 (UnityEngine.Color)
IL_07C9: initobj    UnityEngine.Color
IL_07CF: stloc      27 (UnityEngine.Color)
IL_07D3: ldarg.1
IL_07D4: ldloc      27 (UnityEngine.Color)
IL_07D8: stobj      UnityEngine.Color
IL_07DD: ret
```

**To Reproduce**

1. The original method and its signature and class:

```c#
// public class CommonBuildingAI : BuildingAI
public override Color GetColor(ushort buildingID, ref Building data, InfoManager.InfoMode infoMode)
```
2. The patch (applied with PatchAll):

```c#
    [HarmonyPatch]
    public static class CommonBuildingAI_GetColor {
        public static MethodBase TargetMethod() {

            return typeof(CommonBuildingAI).GetMethod(
                "GetColor",
                BindingFlags.Instance | BindingFlags.Public,
                null,
                new[] { typeof(ushort), typeof(Building).MakeByRefType(), typeof(InfoManager.InfoMode) },
                new ParameterModifier[0]);
        }

        public static bool Prefix(CommonBuildingAI __instance, ushort buildingID, ref Building data, InfoManager.InfoMode infoMode, ref Color __result) {
            __result = Color.red;
            return false;
        }
    }
```

3. The output of the Harmony debug log

[harmony.log.txt](https://github.com/pardeike/Harmony/files/4418309/harmony.log.txt)

4. The stacktrace or other errors

Pretty random errors, mostly NullRefExceptions that occur outside or within the patched method, or "attempted to access invalid address":
```
========== OUTPUTING STACK TRACE ==================

0x0000000004FD1B57 (Mono JIT Code) (wrapper dynamic-method) CommonBuildingAI:DMD<DMD<GetColor_Patch1>?1297982592::GetColor_Patch1> (CommonBuildingAI,intptr,uint16,Building&,InfoManager/InfoMode)
0x0000000070915A70 (Mono JIT Code) PlayerBuildingAI:GetColor (uint16,Building&,InfoManager/InfoMode)
0x00000000708E933A (Mono JIT Code) DepotAI:GetColor (uint16,Building&,InfoManager/InfoMode)
0x000000005492DF47 (Mono JIT Code) BuildingManager:UpdateColorMap (UnityEngine.Texture2D)
0x000000005492DBFB (Mono JIT Code) RenderManager:UpdateColorMap ()
0x000000001A5EC8BD (Mono JIT Code) RenderManager:LateUpdate ()
0x00000000064B050B (Mono JIT Code) (wrapper runtime-invoke) object:runtime_invoke_void__this__ (object,intptr,intptr,intptr)
0x00007FFA70E45CDF (mono) mono_set_defaults
0x00007FFA70D98499 (mono) mono_runtime_invoke
  ERROR: SymGetSymFromAddr64, GetLastError: 'Es wurde versucht, auf eine unzulässige Adresse zuzugreifen.' (Address: 00007FF6BF9E4954)
...
0x00007FF6BFDC6A0C (Cities) 
0x00007FFAC2837BD4 (KERNEL32) BaseThreadInitThunk
0x00007FFAC34ACED1 (ntdll) RtlUserThreadStart

========== END OF STACKTRACE ===========

**** Crash! ****
```

**Runtime environment (please complete the following information):**
 - OS: Windows 10, 64bit
 - .NET version 3.5
 - Harmony version 2.0.0.9
- Name of game or host application: Cities Skylines

boformer:
This is an attempt to fix #277 

For now I just extended `Test_Patch_Returning_Structs` to also validate the returned values.

Unit tests result on my machine are very incoherent. I'm submitting this PR so that the CI can run the updated tests for (hopefully) better results.


boformer:
@pardeike It seems that Harmony behavior with methods that return structs is still very buggy.

When looking at the CI test result, it seems that every version of .NET Framework/.NET Core/Mono handles the `IntPtr` for returning structs differently.

I wonder if it would make sense to change the Harmony code so that it doesn't rely on hardcoded behavior. Instead it could test for each struct size and static/instance configuration if the `IntPtr` fix is needed or not before patching.

pardeike:
Is this still necessary?

PR #283, opened 2020-04-05
Fix issue #275 - Update MMC and add missing InternalsVisibleTo

### LINKED ISSUE #275: The type initializer for 'HarmonyLib.AccessTools' threw an exception
I'm getting this exception while calling PatchAll:

```
System.TypeInitializationException: The type initializer for 'HarmonyLib.AccessTools' threw an exception. ---> System.TypeLoadException: Access is denied: 'MonoMod.Utils.Cil.ILGeneratorShim'. 
```

What does it mean?

0x0ade:
This PR fixes #275 by updating MMC to 20.4.5.1 and adds `[assembly: InternalsVisibleTo("MonoMod.Utils.Cil.ILGeneratorProxy")]` to Harmony's `AssemblyInfo.cs`

### DIFF
--- a/Harmony/Harmony.csproj
+++ b/Harmony/Harmony.csproj
@@ -85,7 +85,7 @@
 
 	<ItemGroup>
 		<!-- https://dev.azure.com/MonoMod/MonoMod/_packaging?_a=feed&feed=DevBuilds -->
-		<PackageReference Include="MonoMod.Common" Version="20.3.29.2">
+		<PackageReference Include="MonoMod.Common" Version="20.4.5.1">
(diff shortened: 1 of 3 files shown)


0x0ade:
[The .NET Framework 4.8 test process seems to crash non-deterministically in some environments.](https://ci.appveyor.com/project/pardeike/harmony/builds/31969365/job/ukbp91tunrv1quy6#L2404) I was able to replicate the crash once on my PC so far, but only during the first run outside of VS without any debugger attached, which means that there isn't even any info as to what could be causing it...

PR #290, opened 2020-04-24
Support netcoreapp2.1

pengweiqhca:
Simplify TargetFrameworks, delete unnecessary reference


pardeike:
If you can make it so all CI tests don’t fail then I might accept it. Right now, you broke 2 out of 3 platforms.

PR #298, opened 2020-05-29
Update doc: Remove nested type example

lucianposton:


pardeike:
Oh, did not see that. Need to fix the other pr too

pardeike:
Just saw this: Your PR changed the compiled result and will be overwritten by the build system. Can you submit a new PR that edits the markdown files in Harmony/Documentation or the xml triple comments in the source files where the method is defined.

Pathoschild:
This was clarified via #300, so we might not another PR for this.

PR #308, opened 2020-06-11
Add AccessTools.MethodDelegate and basic unit tests

lbmaian:
Based off https://gist.github.com/lbmaian/2369d6614a9e56ad2392254b4638d096 and discussion in discord

### DIFF
--- a/Harmony/Tools/AccessTools.cs
+++ b/Harmony/Tools/AccessTools.cs
@@ -968,13 +968,18 @@ namespace HarmonyLib
 				throw new ArgumentNullException(nameof(method));
 
 			var delegateType = typeof(DelegateType);
+
+			// Static method delegate
 			if (method.IsStatic)
 			{
 				return (DelegateType)Delegate.CreateDelegate(delegateType, method);
 			}
-			else if (instance is null)
+
+			// Open instance method delegate ...
+			if (instance is null)
 			{
 				var declaringType = method.DeclaringType;
+				// ... that virtually calls
 				// Note: struct instance methods actually have their internal instance parameter passed by ref,
 				// and thus are incompatible with typical delegates
 				// (delegate type must be: delegate <return type> MyStructDelegate(ref MyStruct instance, ...)),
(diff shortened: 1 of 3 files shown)


Harmony/Tools/AccessTools.cs
```diff
@@ -937,6 +937,102 @@ public static FieldRef<F> StaticFieldRefAccess<F>(FieldInfo fieldInfo)
...
+		/// </para>
+		/// </remarks>
+		public static DelegateType MethodDelegate<DelegateType>(MethodInfo method, object instance = null, bool virtualCall = true)
+			where DelegateType : Delegate
+		{
+			if (method is null)
+				throw new ArgumentNullException(nameof(method));
+
+			var delegateType = typeof(DelegateType);
+			if (method.IsStatic)
```
pardeike, line 971: Using `if() { return }` without `else` makes the whole construct easier to read with less indention

lbmaian: I can remove that `else` but it wouldn't reduce any indentation. Only removing `else` for the `if (instance is null)` would do that, but that's arguably less readable, since it's like (but not exactly):

```cs
if (instance is null)
{
  if (virtualCall)
    ...
  else
    ...
}
else
{
  if (virtualCall)
    ...
  else
    ...
}
```

pardeike: How about this structure:
```
if (static) 
{
   ...
   return
}

if (instance == null)
{
   if (virtual)
   {
       ...
       return
   }
   ...
   return
}

if (virtual) return
...
return
```

pardeike: And maybe switch so instance==null is last

HarmonyTests/Tools/TestAccessTools.cs
```diff
@@ -288,5 +289,97 @@ public void Test_AccessTools_FieldRefAccess_Anonymous()
 			value = newValue;
 			Assert.AreEqual(newValue, fieldInfo.GetValue(instance));
 		}
+
+		[Test]
+		public void Test_AccessTools_MethodDelegate()
```
pardeike, line 294: I like that this does not end up like 100 tests - faster and in this case just fine

Harmony/Tools/AccessTools.cs
lbmaian: Well this where the "but not exactly" comes into play. I'll change it to this:

```cs
if (static)
{
   return ...
}

if (instance is null)
{
   if (virtualCall && !declaringType.IsValueType)
   {
     return ...
   }

   ...
   return ...
}

if (virtualCall)
{
  return ...
}

...
return ...
```

pardeike: Looks fine to me

PR #310, opened 2020-06-12
More AccessTools delegate work

lbmaian:
Support interface method delegates and open instance delegates with interface instance.
Fix mono workaround's inability to access non-public delegate type.
Add more unit tests.


pardeike:
Nice work!

PR #312, opened 2020-06-13
Mark Memory and PatchInfo internal

Pathoschild:
They're used internally and not exposed through any public API.

### DIFF
--- a/Harmony/Internal/Memory.cs
+++ b/Harmony/Internal/Memory.cs
@@ -6,7 +6,7 @@ using System.Runtime.CompilerServices;
 namespace HarmonyLib
 {
 	/// <summary>A low level memory helper</summary>
-	public static class Memory
+	internal static class Memory
 	{
 		/// <summary>Mark method for no inlining (currently only works on Mono)</summary>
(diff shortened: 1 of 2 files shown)


pardeike:
And what is the motivation that justifies the fact that this will:
- confuse beginners
- make it easier to abuse the API
- and ultimately hinders the implementation of changes because public API can no longer be changed without breaking the usage of Harmony (which often is embedded in games and hard to update)

pardeike:
Wait. I misread this. You want to make those classes internal?! Memory is used for detouring when necessary and PatchInfo is definitely used for fetching information about patches

Pathoschild:
Yep, this makes them internal per the reasons in your first comment. Re usage:
* `PatchInfo` is only used internally to track patches for a method. The public API uses `Patch`.
* `Memory` is in the `Harmony\Internal` folder, so I assumed it wasn't meant to be used publicly. I can remove it from the PR if it's part of the public API though.

Pathoschild:
Per discussion, since it's public there's a chance someone was referencing it so this would be a breaking change. I'll close this, and change my upcoming PR to be backwards-compatible just in case.

PR #313, opened 2020-06-14
Reduce array copies in patch tracking

Pathoschild:
When Harmony tracks patches for a method, it copies the patch list twice for each added patch (copy array into a list, add one entry, copy it back into an array, repeat). This commit reduces that to two copies regardless of the number of patches being added, and reduces duplicate code.

In other words, Harmony currently does this to add three patches:
>1. Copy array to list.
>2. Add patch to list.
>3. Copy list to array.
>4. Copy array to list.
>5. Add patch to list.
>6. Copy list to array.
>7. Copy array to list.
>8. Add patch to list.
>9. Copy list to array.

With this change, it works like this instead:
>1. Copy array to list.
>2. Add all new patches to list.
>3. Copy list to array.

Notes:
* `PatchInfo` isn't technically part of the API, but since it's public this keeps backwards compatibility just in case per discussion.
* Since the fields are identical and Harmony uses a type binder, this should have no effect on binary serialization even if multiple Harmony versions are involved.


pardeike:
This would need to be updated to fix the merge conflict and to run the latest CI tests

pardeike:
Great work!

PR #314, opened 2020-06-14
More HarmonyDelegate tweaks

lbmaian:
Edited quote from discord:

### `MethodBinding` => `MethodDispatchType`

I'm having second thoughts on the `MethodBinding` naming, since "binding" is an unfortunately overused term, even within delegates, e.g. bound/unbound delegates. And there's also the existing `System.Reflection.Binder` and `System.Runtime.Serialization.SerializationBinder` and `Microsoft.CSharp.RuntimeBinder.Binder` and `System.Runtime.CompilerServices.CallSiteBinder` classes, and that's not even getting into all the "data binding" stuff.

I think "dispatch" as suggested earlier is a better option, since it's far less overloaded of a term. Indeed dynamic dispatch is considered a specific type of dynamic/late binding (although the difference might be academic in C#, excluding Harmony patching; probably only relevant for languages that can add/remove/change methods at runtime). In fact, "virtual dispatch" isn't the only form of "dynamic dispatch" that C# can use - there's also the `dynamic` type, which is basically a fully dynamic binding (name and overload resolution happens at runtime, method doesn't even need to exist at compile-time).

So `MethodBinding` has been renamed to `MethodDispatchType`, and XML docs updated accordingly.

### HarmonyMethod.virtualDelegate defaulting logic

There's no need for `virtualDelegate` to be a `bool?` (nullable bool) like `HarmonyMethod.debug`. The latter is only nullable, because the default value for it can changed via `Harmony.DEBUG`.

So `HarmonyMethod.virtualDelegate` defaulting logic has been moved into `HarmonyMethod` itself, and XML docs updated accordingly (including what is the default value).

### `virtualCall`/`CallVirtually`/`Call` naming

`MethodDispatchType.CallVirtually` is renamed to `MethodDispatchType.VirtualCall` to:
a) align with `AccessTools.MethodDelegate`'s `virtualCall` parameter
b) be a noun with adjective (like most enum value names), rather than a verb with adverb

#### Open questions

1. Should `AccessTools.MethodDelegate` use a `MethodDispatchType` parameter instead of `bool virtualCall` parameter?

2. Should `MethodDispatchType.Call` be renamed to something like `StaticCall`? I know there were concerns with confusion with static methods, but "Call" just by itself is also kind of confusing.

### DIFF
--- a/Harmony/Internal/MethodPatcher.cs
+++ b/Harmony/Internal/MethodPatcher.cs
@@ -503,7 +503,7 @@ namespace HarmonyLib
 									}
 								}
 
-								if (!methodInfo.IsStatic && harmonyMethod.virtualDelegate)
+								if (methodInfo.IsStatic == false && harmonyMethod.nonVirtualDelegate == false)
 								{
 									emitter.Emit(OpCodes.Dup);
 									emitter.Emit(OpCodes.Ldvirtftn, methodInfo);
--- a/Harmony/Internal/MethodPatcher.cs
+++ b/Harmony/Internal/MethodPatcher.cs
@@ -503,7 +503,7 @@ namespace HarmonyLib
 									}
 								}
 
-								if (!methodInfo.IsStatic && (harmonyMethod.virtualDelegate ?? true))
+								if (!methodInfo.IsStatic && harmonyMethod.virtualDelegate)
 								{
 									emitter.Emit(OpCodes.Dup);
 									emitter.Emit(OpCodes.Ldvirtftn, methodInfo);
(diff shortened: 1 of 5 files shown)


lbmaian:
I've flipped the `virtualDelegate` flag to `nonVirtualDelegate`, so that the default value of the field would align with the default value of the `bool` type (`false`). It shouldn't matter, since old Harmony version code seeing new `HarmonyMethod` with the new flag won't care about that flag, and new Harmony version code seeing old `HarmonyMethod` without the new flag will be properly initializing the field anyway. And `HarmonyMethod` isn't being serialized either.

PR #315, opened 2020-06-21
Unit test attempt to reproduce mono crash when patching empty virtual method (didn't work)

lbmaian:
Just to show my attempt to reproduce this empty virtual method patch crashing on mono issue. Feel free not to merge.

As discussed on discord: https://discordapp.com/channels/131466550938042369/683015847697973333/722843319851286599

Relevant Ludeon threads:
https://ludeon.com/forums/index.php?topic=52325
https://ludeon.com/forums/index.php?topic=52421
https://ludeon.com/forums/index.php?topic=37241

I tried my best to reproduce it in a unit test, running it in a local mono build, but I couldn't reproduce the crash. I strongly suspect it's a Unity mono bug.

### DIFF
--- a/HarmonyTests/Patching/Assets/Specials.cs
+++ b/HarmonyTests/Patching/Assets/Specials.cs
@@ -169,4 +169,25 @@ namespace HarmonyLibTests.Assets
 			yield return new CodeInstruction(OpCodes.Ret);
 		}
 	}
+
+	public class EmptyVirtualMethodClass
+	{
+		public string x = "init";
+
+		[MethodImpl(MethodImplOptions.NoInlining)]
+		public virtual void Method(string y) { }
+	}
+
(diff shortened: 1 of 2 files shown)


lbmaian:
Oh hey, the unit test failed in non-mono net35/netcoreapp3.0/netcoreapp3.1, indicating that an empty virtual method was inlined and thus couldn't be patched? That's a separate issue.

lbmaian:
Related: Zetrith may have discovered why Unity/Mono on Linux x64 may be crashing: https://discordapp.com/channels/131466550938042369/674571535570305060/732548179807174707

Presuming that's the true root cause, reproducing this in a unit test may be extremely difficult (that is, probably not worth the effort).

lbmaian:
The corresponds to #343 and #337. Going to close this for now - new tests are nice, but until a) they work reliably, b) relevant bugs are fixed.

PR #321, opened 2020-07-19
Exhaustively document current AccessTools.*FieldRefAccess behavior in preparation for improvements

lbmaian:
## Summary
* No changes to any existing behavior in `AccessTools`, let alone Harmony proper.
* Move `AccessTools.*FieldRefAccess` unit tests new `TestFieldRefAccess`.
* Expand such tests to exhaustively document current behavior.
* More `AccessTools.Field/Method/Property` unit tests, including tests showing potential NRE on interface types (relevant to `FieldRefAccess` method that uses `AccessTools.Field`, throwing wrong exception).
* Deprecated `FastAccess` getters/setters.
* Migrate Azure Pipelines to use `dotnet` CLI tasks rather than `nuget/vsbuild/vstest` tasks (since the latter was crashing on a unit test that `dotnet test` isn't crashing on).

### Miscellaneous changes
* New `AccessTools.IsStatic`/`IsNetFrameworkRuntime`/`IsNetCoreRuntime` methods.
* Link Mono bug issue with `AccessTools.MethodDelegate` (accidentally included, but it's such a minor change that I kept it in the PR).
* Some new `TestTools` methods and improvements to workaround certain NUnit limitations.
* Some new `Dictionary` extension methods - `internal` to avoid extension method pollution on common BCL types.
* Travis and Azure Pipelines CI `dotnet test` output now includes console stdout logging.

## Details
This in preparation for lbmaian/Harmony#1 (`AccessTools.*FieldRefAccess` improvements, including struct support), including binary compatibility testing. Note that GitHub doesn't support dependent PRs, so that is just a draft PR on my own repo fork - if/once this PR gets merged in, I'll recreate lbmaian/Harmony#1 on this repo.

I might have overengineered it a bit with a custom test suite framework for testing field get/set built on top of NUnit, but the result is a very exhaustive documentation of all the use cases (whether valid or not) of `AccessTools.*FieldRefAccess`.

Some tested use cases cause crashes, namely attempting to use them for struct instance fields (which is the main impetus behind these tests and lbmaian/Harmony#1). These are marked as `[Explicit]` unit tests. (Note: Due to a bug in the NUnit Visual Studio adapter though, `[Explicit]` is ignored in Visual Studio, so I had to add a workaround for this.)

Sample outputs
* Mono 6.10 runtime on debug net472 build: [TestFieldRefAccess-old.mono.net472.log](https://github.com/pardeike/Harmony/files/4945898/TestFieldRefAccess-old.mono.net472.log)
* CLR 4 on release net472 build: [TestFieldRefAccess-old.release.net472.log](https://github.com/pardeike/Harmony/files/4945902/TestFieldRefAccess-old.release.net472.log)
* CoreCLR 3.1 on release netcoreapp3.1 build: [TestFieldRefAccess-old.release.netcoreapp3.1.log](https://github.com/pardeike/Harmony/files/4945909/TestFieldRefAccess-old.release.netcoreapp3.1.log)

Among the commits were my planned changes to `FastAccess` to similarly fix their edge cases, but I've reverted them and instead deprecated the `FastAccess` getters/setters per discord conversation. The reverted commits are there primarily for history (in case my fixes there end up being useful elsewhere). Their remaining legacy is the additional properties/indexers on the `AccessTools` test asset types.


Harmony/Extras/FastAccess.cs
```diff
@@ -154,14 +162,12 @@ public static InstantiationHandler<T> CreateInstantiationHandler<T>()
 			return (SetterHandler<T, S>)dynamicSet.Generate().CreateDelegate(typeof(SetterHandler<T, S>));
 		}
 
-		//
-
-		static DynamicMethodDefinition CreateGetDynamicMethod<T, S>(Type type)
+		private static DynamicMethodDefinition CreateGetDynamicMethod<T, S>(Type type)
```
pardeike, line 165: I think the Harmony code standard does not use `private` and instead let no modifier be implicit private.

Harmony/Tools/AccessTools.cs
```diff
@@ -1156,11 +1157,22 @@ public static void RethrowException(Exception exception)
...
-		/// <returns>True if we are running under Mono, false otherwise (.NET)</returns>
+		/// <summary>True if the current runtime is based on Mono, false otherwise (.NET)</summary>
 		///
 		public static bool IsMonoRuntime { get; } = Type.GetType("Mono.Runtime") != null;
 
+		/// <summary>True if the current runtime is .NET Framework, false otherwise (.NET Core or Mono, although latter isn't guaranteed)</summary>
+		///
+		public static bool IsNetFrameworkRuntime { get; } =
+			TypeByName("System.Runtime.InteropServices.RuntimeInformation")?.GetProperty("FrameworkDescription")
+			.GetValue(null, null).ToString().StartsWith(".NET Framework") ?? true;
```
pardeike, line 1168: The default is `true`? Is that correct?

Harmony/Tools/AccessTools.cs
```diff
@@ -1395,6 +1407,58 @@ public static bool IsOfNullableType<T>(T instance)
 			return Nullable.GetUnderlyingType(typeof(T)) != null;
 		}
 
+		/// <summary>Tests whether a type or member is static, as defined in C#</summary>
+		/// <param name="member">type or member</param>
+		/// <returns>True if the type or member is static</returns>
+		///
+		public static bool IsStatic(MemberInfo member)
```
pardeike, line 1414: Maybe add null check and make it an extension method?

Harmony/Tools/AccessTools.cs
```diff
@@ -1395,6 +1407,58 @@ public static bool IsOfNullableType<T>(T instance)
...
+				default:
+					throw new ArgumentException($"Unknown member type: {member.MemberType}");
+			}
+		}
+
+		/// <summary>Tests whether a type is static, as defined in C#</summary>
+		/// <param name="type">type</param>
+		/// <returns>True if the type is static</returns>
+		///
+		public static bool IsStatic(Type type)
```
pardeike, line 1439: Same as other IsStatic

Harmony/Tools/AccessTools.cs
```diff
@@ -1395,6 +1407,58 @@ public static bool IsOfNullableType<T>(T instance)
...
+		{
+			return type.IsAbstract && type.IsSealed;
+		}
+
+		/// <summary>Tests whether a property is static, as defined in C#</summary>
+		/// <param name="property">property</param>
+		/// <returns>True if the property is static</returns>
+		///
+		public static bool IsStatic(PropertyInfo property)
```
pardeike, line 1448: Same

Harmony/Tools/AccessTools.cs
```diff
@@ -1395,6 +1407,58 @@ public static bool IsOfNullableType<T>(T instance)
...
+		{
+			return property.GetAccessors(true)[0].IsStatic;
+		}
+
+		/// <summary>Tests whether an event is static, as defined in C#</summary>
+		/// <param name="event">event</param>
+		/// <returns>True if the event is static</returns>
+		///
+		public static bool IsStatic(EventInfo @event)
```
pardeike, line 1457: I try to avoid `@`

Harmony/Extras/FastAccess.cs
lbmaian: Fixed in all changed files, except in the asset classes, where I think the explicit `private` makes sense (some other asset classes also use explicit `private`).

Harmony/Tools/AccessTools.cs
lbmaian: Yes, because `System.Runtime.InteropServices.RuntimeInformation` only exists on .NET Core (any version) and .NET Framework 4.7.1+, so if it's missing, it should mean that some old .NET Framework runtime is being used.

Although, it's possible that an old Mono runtime (like that used in RimWorld 1.0 or other old Unity-based versions) also don't have `RuntimeInformation`, so I'll fix it so that that `true` is `IsMonoRuntime is false`.

Harmony/Tools/AccessTools.cs
lbmaian: Added `null` check. I thought about making it an extension method, but it would look awkward with the existing `IsStatic` property on `MethodBase` and `FieldInfo`. If it were possible to add extension properties, I would have gone with that approach.

Harmony/Tools/AccessTools.cs
lbmaian: See above

Harmony/Tools/AccessTools.cs
lbmaian: See above

Harmony/Tools/AccessTools.cs
lbmaian: Changed to `eventInfo`.

PR #323, opened 2020-07-22
Add AccessTools.TypeByName tests

lbmaian:
## `AccessTools.TypeByName`
* Add `AccessTools.TypeByName` tests
  * Uses test `AppDomain`s in .NET Framework
  * Uses test `AssemblyLoadContext`s in .NET Core
* ~Add new "DummyAssembly" projects~ Dynamically generated assemblies to setup an "invalid assembly":
  * `HarmonyTestsDummyAssemblyA` is dependency of `HarmonyTestsDummyAssemblyB` yet `HarmonyTests` explicitly doesn't include `HarmonyTestsDummyAssemblyA`
  * `HarmonyTestsDummyAssemblyB` depends on `HarmonyTestsDummyAssemblyA` and exists in `HarmonyTests` - since `HarmonyTestsDummyAssemblyA` isn't included in `HarmonyTests`, `HarmonyTestsDummyAssemblyB` becomes an "invalid assembly"
  * `HarmonyTestsDummyAssemblyC` exists as a valid assembly loaded in `HarmonyTests` after the invalid assembly `HarmonyTestsDummyAssemblyB`
* Add TODOs to `AccessTools.GetTypesFromAssembly` for working around the invalid assembly issue

Functional changes to `AccessTools.TypeByName/GetTypesFromAssembly` that workaround the invalid assembly issue will be done in a separate PR.

## Misc changes
* ~Adds solution folders in an attempt to avoid a cluttered solution project view.~
* ~And since solution folders always appear above root projects in VS, all root projects have been moved into solution folders so that test projects still appear below other projects.~
* ~Add misc solution items like README.md.~
* ~Consolidate/update `jnm2.ReferenceAssemblies.net35` reference version.~
~[2020-07-24 01_11_16-Harmony - Microsoft Visual Studio](https://user-images.githubusercontent.com/30856860/88372431-bd92a600-cd4a-11ea-9922-271d7b693bf2.png)~
* Add dependency on `Lokad.ILPack` for .NET Core to workaround lack of `AssemblyBuilder.Save` support.
* Add `TestTools.RunInIsolationContext` and `ITestIsolationContext` that can execute tests in a separate `AppDomain` (if .NET Framework) or separate `AssemblyLoadContext` that's unloaded afterwards.

~I'm not particularly happy about needing "dummy assembly" projects just to generate the invalid assembly test case, but my attempts so far at dynamically generating them via `AssemblyBuilder` has run into blockers (in particular, .NET Core lacks `AssemblyBuilder.Save`).~


Harmony/Tools/AccessTools.cs
```diff
@@ -56,6 +56,14 @@ public static Type TypeByName(string name)
 		/// 
 		public static Type[] GetTypesFromAssembly(Assembly assembly)
```
pardeike, line 57: As discussed earlier, the best option is to only catch `ReflectionTypeLoadException` and use the valid types for the result. Keep throwing any other exception.

Also refactor any code that uses `assembly.GetTypes()` to use this method.

lbmaian: Alright, I'll do this in the followup PR to this (the one that actually does the fixes).

lbmaian:
Fixed merge conflicts and fixed .editorconfig not actually applying code style rules. Will merge once CI finishes.

PR #328, opened 2020-07-28
Misc solution/project/test/code cleanup and fixes, CI fully migrated to Azure Pipelines

lbmaian:
## Solution/project config cleanup and improvements
- Add misc solution items (like README.md) to solution
- Allow C# 8
- More robust Zip build task that works on .NET Core dotnet build (fixes non-Windows release builds)
  - Migrate Zip build task to use RoslynCodeTaskFactory (available for msbuild versions since VS 2017 15.8+ and .NET Core 2.1+)
  - All old *.zip files in output dir are deleted upon build and clean
- Dependency updates:
  - Consolidate/update `jnm2.ReferenceAssemblies.net35` reference version (fixes the 'warning MSB3245: Could not resolve this reference. Could not locate the assembly "System.Xml"' on non-Windows builds)
  - Update `Microsoft.NET.Test.Sdk` and `NUnit3TestAdapter` dependencies
  - Remove unused netstandard2.0/netcoreapp2.0 dependencies and already commented out dependencies
  - Remove libs folder that was only used for netstandard2.0
  - Remove unused test dependencies in Documentation project
- Future-proof csproj .NET Core conditions to handle .NET 5+
- Add global.json to prevent NET 5 pre-release from being used via dotnet CLI commands, since Harmony dependencies like Cecil aren't compatible with it yet
- Misc updates to .editorconfig to match existing code style, and generalize yml settings for all yml and json.

## Continuous integration revamp (full migration to Azure Pipelines)
- Remove travis.yml and appyveyor.yml
- Revamp Azure Pipelines CI config to test across dotnet/mono win-x64/win-x86/ubuntu-x64/macOS-x64
  - ubuntu-x86 and macOS-x86 currently not possible without x86 hosts
  - Note that ubuntu and macOS implicitly uses mono for .NET Framework and thus there aren't separate mono jobs for them
  - There was an attempt to attach stdout/stderr to non-failed test results (Azure Pipelines only attaches such for failed tests), but it was reverted due to significantly slowing down test result publishing (see commit history)
- A workaround for running net35 tests via `dotnet test` without having to specify dll (unifies how tests are run across all target frameworks)
- Publishes the release Harmony zip as build artifact

## Test improvements
- Attempt to improve test stability by catching CSEs in ReturningStructs tests, which occasionally causes crashes, in .NET Framework 4+
  - Ideally, these tests would each be done in separate process to isolate crashes, but there's no way easy way to do that (I've been experimenting with a solution, but it's .NET Core-specific)
- Add TestEnvironment "tests" that logs runtime/environment info when running tests
- Made `[Explicit]` VSTest workaround automatic and fixed to work under mono and work with vstest.console
- Made test class namespaces consistent with their folders (some already had such namespaces, so this is a consistency change)

## Code cleanup
- Fix some formatting and code style issues (address suggestions from Visual Studio)
- Use `is` over `==`/`!=` for `null`/`false` checks - more idiomatic in modern C# and marginally improves performance for types that implement equality operators (such as Type, FieldInfo, etc.)
  - `!= null` is replaced with `is object` for now as the least bad option (should be changed to `is not null` when C# 9 is released)
- Add EditorBrowsableState.Never to the specific AccessTools.IsStatic methods so that only MemberInfo overload should show up in IDEs by default


pardeike:
What a monster. 85 files changed/touched

PR #331, opened 2020-08-09
Patch static generic methods

abrenneke:
This adds support for patching generic methods. Instead of `Method<string>` being patched and all `Method<referenceType>` becoming `Method<string>`, `Method<>` is patched, and then the runtime can generate generics using `MakeGenericMethod`.

Currently, static generic methods of non-generic classes can be patched, and maintain their type information. 

Requires the [MonoMod.Common changes here](https://github.com/MonoMod/MonoMod.Common/pull/8).

### DIFF
--- a/Harmony/Harmony.csproj
+++ b/Harmony/Harmony.csproj
@@ -25,6 +25,7 @@
 		<AllowUnsafeBlocks>true</AllowUnsafeBlocks>
 		<LangVersion>7.3</LangVersion>
 		<DefaultItemExcludes>$(DefaultItemExcludes);Documentation/**</DefaultItemExcludes>
+		<AutoGenerateBindingRedirects>true</AutoGenerateBindingRedirects>
 	</PropertyGroup>
 
 	<Choose>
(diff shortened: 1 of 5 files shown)


drok:
@pardeike, are you planning on finishing and integrating the work that @abrenneke started into Harmony?

If not, would you please outline what is needed to integrate it, so that I or someone else can tackle it?

I desperately need to patch generic members using generic patch functions, to work with generic parameters that are not known at patch time. The list of generic parameters on the patch function needs to be a subset of the list of generics on the original function, just like the method parameters, ie `bool Prefix<T1, T2...>(Type1 par1, Type 2 par2...)`

I am willing to do the work, but lack of documentation for MonoMod.Utils and Harmony internals are big stumbling blocks, as it steepens the learning curve. I'm hoping that your outline will give me enough clues as to get started. I'm able to learn from documentation, source code and debugger session, but I'm at the beginning of the learning curve, given that I didn't even know runtime patching assemblies was possible until a month or two ago.

Thank you!

PR #333, opened 2020-08-21
Fix a DirectoryNotFoundException at FileLog

iRebbok:
This PR fixes a `DirectoryNotFoundException` at FileLog when it tries to log into a file on a docker container. `Environment.GetFolderPath(Environment.SpecialFolder.Desktop)` returns `/home/container/Desktop`, but this directory doesn't exists by default, and it causes a `DirectoryNotFoundException`.


pardeike:
Fixed in master with a better implementation

PR #336, opened 2020-08-22
Fix build: migrate from ILRepack.Lib.MSBuild.Task to ILRepack.Lib.MSBuild

lbmaian:
#335 included migrating from `ILRepack.MSBuild.Task` to `ILRepack.Lib.MSBuild.Task`, since the former is no longer maintained (its GitHub repo has been archived) and the latter drastically reduces logging that included warnings on non-Windows.

However, although the project built locally in Visual Studio, it broke `dotnet` builds due to `ILRepack.Lib.MSBuild.Task`'s usage of outdated `Microsoft.Build.Utilities.v4.0`.

I found a more maintained ILRepack build task that does work with `dotnet`, yet keeps the benefit of reduced logging:  `ILRepack.Lib.MSBuild`. So this migrates from `ILRepack.Lib.MSBuild.Task` to `ILRepack.Lib.MSBuild`.

As a bonus, `ILRepack.Lib.MSBuild` is both easier to use and uses a more maintained fork of `ILRepack`/`ILRepack.Lib`.


lbmaian:
I had to workaround a problem with macOS builds requiring a DYLD_* env var to be set (since it's unset likely due to https://developer.apple.com/library/archive/documentation/Security/Conceptual/System_Integrity_Protection_Guide/RuntimeProtections/RuntimeProtections.html).

Harmony/Harmony.csproj
```diff
@@ -64,17 +97,46 @@
 	<ItemGroup>
 		<!-- https://dev.azure.com/MonoMod/MonoMod/_packaging?_a=feed&feed=DevBuilds -->
 		<PackageReference Include="MonoMod.Common" Version="20.5.31.1">
-			<PrivateAssets Condition="!$(IsCoreOrStandard) And '$(Configuration)'=='Release'">All</PrivateAssets>
+			<PrivateAssets Condition="$(ILRepackInternalize)">all</PrivateAssets>
```
pardeike, line 100: Nice with an extra variable here!

pardeike (approved):

pardeike:
Since it works, can we merge to master and fix subsequent issues in a new PR?

lbmaian:
Sure. Only change I've made since is just updating some comments.

lbmaian:
There will be another followup PR since it turns out `ILRepack.Lib.MSBuild` is also problematic. It uses a fork of ILRepack that changes the internalization algorithm that results in some of non-Harmony types no longer being internalized.

PR #347, opened 2020-09-12
More thread-safety

lbmaian:
`HarmonySharedState`:
- A lock on a `const` string doesn't look safe (might be safe due to the interned string, but let's not bank on it), so locking on an explicit lock object.
- Also locking over the whole operation, rather than just on `GetState`.

`PatchTools`:
- Attributed `objectReferences` with `ThreadStatic`. Even though this Dictionary is only stored to and never read from, it still needs to be thread-safe: https://stackoverflow.com/a/33153868

`AccessTools`:
- Make `AccessTools.all`/`allDeclared` `static readonly`, with a note on why they're not `const` (would break binary compatibility). I'm making the probably safe assumption that no one ever sets these fields.

Other changes:
- Revert `StructReturnBuffer.sizes` back to `static readonly` now that tests are clearing it rather than setting it to a new dictionary.
- Add locking to `FileLog.ChangeIndent` - for consistency with other side-effecting methods here.

### DIFF
--- a/Harmony/Internal/HarmonySharedState.cs
+++ b/Harmony/Internal/HarmonySharedState.cs
@@ -10,33 +10,31 @@ namespace HarmonyLib
 	internal static class HarmonySharedState
 	{
 		const string name = "HarmonySharedState";
+		static readonly object locker = new object();
 		internal const int internalVersion = 100;
 		internal static int actualVersion = -1;
 
 		static Dictionary<MethodBase, byte[]> GetState()
 		{
-			lock (name)
+			var assembly = SharedStateAssembly();
+			if (assembly is null)
 			{
-				var assembly = SharedStateAssembly();
-				if (assembly is null)
-				{
-					CreateModule();
-					assembly = SharedStateAssembly();
-					if (assembly is null) throw new Exception("Cannot find or create harmony shared state");
-				}
-
-				var type = assembly.GetType(name);
-
-				var versionField = type.GetField("version");
-				if (versionField is null) throw new Exception("Cannot find harmony state version field");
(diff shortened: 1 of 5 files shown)


lbmaian:
Well it didn't prevent `Test_Patch_Returning_Structs(20,"S")` test failing spuriously: https://pardeike.visualstudio.com/Harmony/_build/results?buildId=1015&view=logs&j=c9daabb1-7301-5a00-f34f-3cec5819fa71&t=9b344549-9f9a-5df6-f54c-b2e622c743be

But I wasn't holding hopes that it would. That test's stability seems only tangentially related to thread safety.

pardeike:
Regarding changing HarmonySharedState: what happens when two versions of Harmony with different lock objects try to access the state at the same time? I think it breaks the backwards compatibility and we need to be careful with the change.

lbmaian:
Hmm, isn't that already possible before? Two versions of Harmony already have their own locks, and each just locks on `GetState()`. This change only expands those same assembly-specific locks (kinda, replace string with locker) over the subsequent operation on the returned state. So it should still be backwards compatible.

Though now that I think of it, this is insufficient. The returned state dictionary itself needs to be locked. And I can "revert" to have the assembly-specific locks only lock in `GetState()` again.

That change should also still be backwards compatible. Old versions that aren't aware of the locking will just ignore the lock (locking requires explicit cooperation after all).

lbmaian:
Also just realized that `lock (state) return state.Keys` doesn't provide any thread-safety. You'd have to either resolve it fully into an array or list, or use `ConcurrentDictionary` (which isn't available in net35 and isn't backwards compatible, although the latter concern can be addressed by versioning).

PR #348, opened 2020-09-25
Add handler cache for MakeDeepCopy

Mhburg:
The following image is a screenshot from profiling the patch time in RogueTech, a heavily modded version of BattleTech.
As shown in the pic, `MakeDeepCopy` accounts for more than 90% of cpu time. The majority of the time is spent on getting the handler of `Add` method for types.
This PR is a simple optimization on this process by creating a dictionary for fast repetitive accesses.
image

This is the profiling data after the change.
image

Following is a comparison on memory usage,
Before,
image

After,
image

The difference mostly can attribute to transient memory allocation, and as the memory usage shown in the After change, the allocation for persistent memory is quite small. However, it would not hurt if Harmony can open an API for modloaders to clear the cache.

### DIFF
--- a/Harmony/Tools/AccessTools.cs
+++ b/Harmony/Tools/AccessTools.cs
@@ -8,6 +8,7 @@ using System.Linq;
 using System.Reflection;
 using System.Reflection.Emit;
 using System.Runtime.Serialization;
+using System.Threading;
 
 namespace HarmonyLib
 {
@@ -15,6 +16,13 @@ namespace HarmonyLib
 	///
 	public static class AccessTools
 	{
+		/// <summary>
+		/// A cache for the <see cref="ICollection{T}.Add"/> or similar Add methods for different types.
+		/// </summary>
+		private static readonly Dictionary<Type, FastInvokeHandler> handlerCache = new Dictionary<Type, FastInvokeHandler>();
+
+		private static readonly ReaderWriterLockSlim handlerCacheLock = new ReaderWriterLockSlim(LockRecursionPolicy.SupportsRecursion);
+
 		/// <summary>Shortcut for <see cref="BindingFlags"/> to simplify the use of reflections and make it work for any access level</summary>
 		///
 		// Note: This should a be const, but changing from static (readonly) to const breaks binary compatibility.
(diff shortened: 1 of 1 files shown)


Mhburg:
Forgot to write, Harmony 1.2 is used in the profiling.
Enviroment is Win10 mono x64

Harmony/Tools/AccessTools.cs
```diff
@@ -1703,9 +1715,6 @@ public static object MakeDeepCopy(object source, Type resultType, Func<string, T
 					}
 					return addableResult;
 				}
-
-				// TODO: add dictionaries support
```
lbmaian, line 1707: Does this already handle dictionaries then via `Add(KeyValuePair<...>)`?

Harmony/Tools/AccessTools.cs
```diff
@@ -1687,11 +1692,18 @@ public static object MakeDeepCopy(object source, Type resultType, Func<string, T
 
 			if (type.IsGenericType && resultType.IsGenericType)
 			{
-				var addOperation = FirstMethod(resultType, m => m.Name == "Add" && m.GetParameters().Count() == 1);
-				if (addOperation is object)
+				if (!handlerCache.TryGetValue(resultType, out FastInvokeHandler addInvoker))
+				{
+					var addOperation = FirstMethod(resultType, m => m.Name == "Add" && m.GetParameters().Count() == 1);
```
lbmaian, line 1697: Sneak in another minor optimization here: use `Count` rather than `Count()`

Harmony/Tools/AccessTools.cs
```diff
@@ -15,6 +15,11 @@ namespace HarmonyLib
 	///
 	public static class AccessTools
 	{
+		/// <summary>
+		/// A cache for the <see cref="ICollection{T}.Add"/> or similar Add methods for different types.
+		/// </summary>
+		private static readonly Dictionary<Type, FastInvokeHandler> handlerCache = new Dictionary<Type, FastInvokeHandler>();
```
lbmaian, line 21: This isn't thread-safe. Ideally we'd use a ConcurrentDictionary, but that isn't supported on net35.

In other areas, we've been using locks as a workaround, but if performance is a concern, then there would have to be two code paths:

net35: uses Dictionary and either:
a) locks on that Dictionary upon usage
b) use ThreadStatic (not ideal for a cache) with manual per-thread initialization

Other: use ConcurrentDictionary

That said, I think always locking is fine, since the dict access performance is minor compared to the addInvoker usage.

lbmaian (changes requested):
Harmony CI is having spurious failures again... I'll keep retrying it manually.

Harmony/Tools/AccessTools.cs
Mhburg: Good to hear that Harmony is moving to multi-thread

Mhburg: I will switch the code to ConcurrentDictionary

lbmaian: You can't use ConcurrentDictionary on .NET Framework 3.5 though, which Harmony must still support. This is why I suggest just locking instead.

Harmony/Tools/AccessTools.cs
lbmaian: Or rather, `Length` since `GetParameters()` returns an array.

Mhburg:
Is preprocessor allowed in Harmony? I can use them for difference version of .NET

On Sep 25, 2020, at 7:56 AM, lbmaian <notifications@github.com> wrote:

﻿

@lbmaian commented on this pull request.

________________________________

In Harmony/Tools/AccessTools.cs<https://github.com/pardeike/Harmony/pull/348#discussion_r494937590>:

> @@ -1687,11 +1692,18 @@ public static object MakeDeepCopy(object source, Type resultType, Func<string, T

                        if (type.IsGenericType && resultType.IsGenericType)
                        {
-                               var addOperation = FirstMethod(resultType, m => m.Name == "Add" && m.GetParameters().Count() == 1);
-                               if (addOperation is object)
+                               if (!handlerCache.TryGetValue(resultType, out FastInvokeHandler addInvoker))
+                               {
+                                       var addOperation = FirstMethod(resultType, m => m.Name == "Add" && m.GetParameters().Count() == 1);


Or rather, Length since GetParameters() returns an array.

lbmaian:
> Is preprocessor allowed in Harmony? I can use them for difference version of .NET

Yes, `#if NET35` blocks already exist in AccessTools.cs. That said, @pardeike wasn't in favor of #if NET35` lock blocks due to code clarity concerns, and decided to just always use locks.

Mhburg:
> > Is preprocessor allowed in Harmony? I can use them for difference version of .NET
> 
> Yes, `#if NET35` blocks already exist in AccessTools.cs. That said, @pardeike wasn't in favor of `#if NET35` lock blocks due to code clarity concerns, and decided to just always use locks.

ok

Harmony/Tools/AccessTools.cs
Mhburg: It is legacy code, I will make the change.

Harmony/Tools/AccessTools.cs
Mhburg: that's a legacy comment from Pardeike

Harmony/Tools/AccessTools.cs
```diff
@@ -1687,25 +1695,45 @@ public static object MakeDeepCopy(object source, Type resultType, Func<string, T
...
+						return addableResult;
 					}
-					return addableResult;
 				}
-
-				// TODO: add dictionaries support
-				// maybe use methods in Dictionary<KeyValuePair<TKey,TVal>>
+				finally
+				{
+					handlerCacheLock.ExitReadLock();
```
lbmaian, line 1735: Shouldn't this be `ExitUpgradeableReadLock`?

lbmaian (changes requested):

lbmaian (approved):
Looks good, will wait for CI to succeed before merging.

PR #354, opened 2020-10-15
Three small fixes and improvements

fbondia:
Hello, it`s my first pull-request and I hope I am doing everything right... Tks

**1 - fix situations where the variable "__result" was declared as "Object" in the "Postfix" method, but the return from the original method is a value-type variable**

In this case I just added an opcode to make the boxing of the value:

```
if (patchParam.ParameterType==typeof(object) && returnType.IsValueType) {
    emitter.Emit(OpCodes.Box, returnType);
}
```

**2 - fix a small typo (path instead of patch)**

There were a text "You can only patch implemented methods/constructors. **Path** the declared method {declaredMember.FullDescription()} instead."

**3 - a prefix or postfix method can have a new special variable "Object[] __args" which receive all the original method parameters, so we can use it as the example bellow**

```

        public static void MyPrefix(Object __instance, MethodBase __originalMethod, Object[] __args)
        {

            Console.WriteLine();
            Console.WriteLine("***********************************");
            Console.WriteLine("** Trampoline");

            if (__args!=null)
            {

                for(int i=0; i<__args.Length; i++)
                {
                    if (__args[i] == null)
                    {
                        Console.WriteLine($"** ARG[{i}]: null");
                    }
                    else {
                        String value = (__args[i] is String ? $"\"{__args[i]}\"" : __args[i].ToString());
                        String type = __args[i].GetType().Name;

                        Console.WriteLine($"** ARG[{i}]: {value} <{type}>");
                    }
                }


            }

        }

```


pardeike:
Hi, thanks for the contribution. I appreciated it although it would be nice if you could:
- remove unnecessary blank lines
- respect `.editorconfig` (`var xxx` instead of `int xxx`)
- submit your changes in separate PR's because its easier to accept them if one of them is not ready yet

I have a few questions about the use cases (1) and (3):

(1) I created the following test case and it works. So I am not sure if this change is necessary:
```cs
public class Class20
{
	public struct Struct20
	{
		public int value;
	}

	public static Struct20 Method20()
	{
		return new Struct20 { value = 123 };
	}
}

[HarmonyPatch(typeof(Class20))]
[HarmonyPatch(nameof(Class20.Method20))]
public static class Class20Patch
{
	public static object theResult;

	public static void Postfix(object __result)
	{
		theResult = __result;
	}
}

[Test]
public void Test_BoxingValueResults()
{
	var struct1 = Class20.Method20();
	Assert.AreEqual(123, struct1.value);

	var harmony = new Harmony("test");
	var processor = new PatchClassProcessor(harmony, typeof(Class20Patch));
	var patches = processor.Patch();
	Assert.NotNull(patches, "patches");
	Assert.AreEqual(1, patches.Count);

	_ = Class20.Method20();
	var result = (Class20.Struct20)Class20Patch.theResult;
	Assert.AreEqual(123, result.value);
}
```

(3) The main two risks with adding this functionality to Harmony is that
a) users will automatically assume you can update arguments by assigning values to array elements
b) edge cases like
- arguments with modifiers: `out string stuff`
- arguments with value references: `ref SomeStruct stuff`
- maybe even `dynamic stuff`

pardeike:
A more careful inspection shows that fixing the case where a patch has a `ref object __result` argument requires quite some detailed handling as this LINQPad example shows:

```cs
void Main()
{
	var x = Class20.Method20();
	Console.WriteLine("" + x.value);
}

public class Class20
{
	public struct Struct20
	{
		public int value;
	}

	public static Struct20 Method20()
	{
		var result = new Struct20 { value = 123 };
		
		Console.WriteLine("START");
		object TEMP = result;
		Patch(ref TEMP); // Patch(ref result) gives a cast error
		result = (Struct20)TEMP;
		Console.WriteLine("END");
		
		return result;
	}
	
	public static void Patch(ref object res)
	{
		res = new Struct20 { value = 456 };
	}
}
```

which gives this IL for the simulated patch call:
```
Class20.Method20:
IL_0000:  ldloca.s    01 
IL_0002:  initobj     UserQuery+Class20.Struct20
IL_0008:  ldloca.s    01 
IL_000A:  ldc.i4.s    7B 
IL_000C:  stfld       UserQuery+Class20+Struct20.value
IL_0011:  ldloc.1     
IL_0012:  ldstr       "START"
IL_0017:  call        System.Console.WriteLine
IL_001C:  box         UserQuery+Class20.Struct20
IL_0021:  stloc.0     // TEMP
IL_0022:  ldloca.s    00 // TEMP
IL_0024:  call        UserQuery+Class20.Patch
IL_0029:  ldloc.0     // TEMP
IL_002A:  unbox.any   UserQuery+Class20.Struct20
IL_002F:  ldstr       "END"
IL_0034:  call        System.Console.WriteLine
IL_0039:  ret
```

fbondia:
Hi, thanks for replying.

I'll split the issues on different PRs, remove the blank lines and take a look at the .editorconfig. I am not used to contribute to open-source projects, so thank you for understanding.

About the boxing of the valuetype returned, I had reproduced it doing this code bellow. I am working on a windows today with .net framework 472, but yesterday was on a mac/mono and the error were also present:

```csharp
public class SimpleClass
    {
        public SimpleStruct SimpleMethod(SimpleStruct s)
        {
            return s;
        }

        public int OtherSimpleMethod()
        {
            return 10;
        }
    }

    public struct SimpleStruct
    {
        public Int32 aInt { get; set; }
        public String aString { get; set; }
    }

    public class Program
    {

        public static void Main(string[] args)
        {

            var prefix = SymbolExtensions.GetMethodInfo(() => MyPrefix(null, null, null, null));
            var postfix = SymbolExtensions.GetMethodInfo(() => MyPostfix(null, null, null, null));

            var target1 = typeof(SimpleClass).GetMethod("OtherSimpleMethod");
            var target2 = typeof(SimpleClass).GetMethod("SimpleMethod");

            var replaced1 = new Harmony($"my.patch").Patch(target1, new HarmonyMethod(prefix), new HarmonyMethod(postfix));
            var replaced2 = new Harmony($"my.patch").Patch(target2, new HarmonyMethod(prefix), new HarmonyMethod(postfix));


            SimpleClass myInstance = new SimpleClass();
            SimpleStruct myStruct = new SimpleStruct() { aInt = 10, aString = "hey" };

            var result1 = myInstance.OtherSimpleMethod();
            /*
             System.NullReferenceException: 'Object reference not set to an instance of an object.'
            __result was null.
             */

            var result2 = myInstance.SimpleMethod(myStruct);
            /*
            Managed Debugging Assistant 'FatalExecutionEngineError' : 'The runtime has encountered a fatal error. 
            The address of the error was at 0x4756f9d0, on thread 0x494. The error code is 0xc0000005. This error may 
            be a bug in the CLR or in the unsafe or non-verifiable portions of user code. Common sources of this bug 
            include user marshaling errors for COM-interop or PInvoke, which may corrupt the stack.'
            */

        }

        public static void MyPrefix(Object __instance, MethodBase __originalMethod, Object __result, Object[] __args)
        {
            Debug.WriteLine("Prefix is not a problem");
        }

        public static void MyPostfix(Object __instance, MethodBase __originalMethod, Object __result, Object[] __args)
        {
            if (__result == null)
            {
                Debug.WriteLine($"** RETURN: null");
            }
            else
            {
                Debug.WriteLine($"** RETURN: \"{__result}\" <{__result.GetType().Name}>");
            }
        }
    }
```

About the other item (3), I understand your concerns. In fact I didnt think about modifying the parameter values. I was only thinking in a read-only use. Make the variables writable will add some complexity for sure... I will take a look at this. Anyway, I think it is a nice feature if it work all right, no?

Anyway, I will split the PRs, make the commented changes, so if you want you can close this one.


Tks

pardeike:
Regarding the boxing, I took the liberty and used your two lines and completed them with some extra cases and commited that to master after writing some tests for the code (btw, always good to add tests to your PR).

Hope you don't mind. Please fetch latest master and remove the redudant lines in your PR.

pardeike:
Here is what I used for reference:
```
// calling static void Patch(xxxx __result), where xxxx has different cases:

// PASSING VALUE TO OBJECT
ldloc.0 // result
box StructType
call Patch

// PASSING VALUE TO REF OBJECT
ldloc.0 // result
box StructType
stloc.1 // temp
ldloca.s 01 // temp
call Patch
ldloc.1 // temp
unbox.any StructType
stloc.0 // result

// PASSING OBJECT TO REF OBJECT
ldloca.s 00 // result
call Patch
```

PR #367, opened 2020-12-07
Update MonoMod.Common to 20.12.7.4

AndreiSingeorzan:
This is so net5 patching works properly.


pardeike:
Thanks for the contribution

PR #379, opened 2021-01-16
Update patching-edgecases.md

podostro:
Added edge case "MissingMethodException: Attempted to access a missing method"

### DIFF
--- a/Harmony/Documentation/articles/patching-edgecases.md
+++ b/Harmony/Documentation/articles/patching-edgecases.md
@@ -72,3 +72,80 @@ public SomeType MyMethod()
 ```
 
 That method has no `RET` IL code in its body and if you try to patch it, Harmony will generate illegal IL. The only solution to this is to create a `Transpiler` that transpiles the method to a correct version by creating valid IL. This is also true for adding a Prefix or Postfix to that method. The way Harmony works, the replacement method needs to be valid to add calls to your patches to it.
+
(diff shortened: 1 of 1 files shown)


pardeike:
After seeing this I wonder if this is too off topic for Harmony. If I would add all the cases similar to this one, the documentation would get bloated and not consumable anymore. 

As a compromise, I am open to add a new chapter called "Unity" or "Runtime Resolving Problems" to the documentation and we move your new section into that. 

What do you think?

podostro:
I actually think this would be a basic guideline to using Harmony for Unity Games, since the intro makes it look like that is the purpose of the library. 
But if your focus is on patching anything and no longer explicitly Unity games; then I aggree, a new Unity focused section would be cleaner.

pardeike:
Sorry for not getting back to you earlier. I think your motivation is good and the only thing I would like you to do is to move the code example out of the md file with a reference. Check out how the rest of the documenation is done.

Once that's fixed I pull this request.

podostro:
The thing with motivations is, it dies... 
Also I'm currently too busy with my own work and not have time to contribute; so if you would be so kind as to do it yourself; or open request anyone else to do it. It would be great. 
Otherwise, time will tell when I'm able to do so.

pardeike:
No worries. I just want to know your status. We all have stuff to do that gets in the way. Thanks for your contribution so far.

PR #397, opened 2021-04-19
Provide contravariance on FieldRef delegate

Therzok:
I'm not 100% sure, but seems like ABI compat is preserved.

Testing out an app that was compiled against a older version of a lib with no contravariance, then running it against a newer version with added contravariance seems to work.

### DIFF
--- a/Harmony/Tools/AccessTools.cs
+++ b/Harmony/Tools/AccessTools.cs
@@ -836,7 +836,7 @@ namespace HarmonyLib
 		/// </para>
 		/// </remarks>
 		///
-		public delegate ref F FieldRef<T, F>(T instance = default);
+		public delegate ref F FieldRef<in T, F>(T instance = default);
 
 		/// <summary>Creates a field reference delegate for an instance field of a class</summary>
 		/// <typeparam name="T">The class that defines the instance field, or derived class of this type</typeparam>


Therzok:
The tests passed locally?

pardeike:
Can you explain what the addition of "in" makes for a difference? Also: I assume your ABI test was done by making sure the new version was the one available at runtime, right?

I reran the failed tests as Azure unfortunately has this kind of randomly failing tests.

Therzok:
Yup, ABI test was done ensuring the right one is loaded (as in the newer version).

The addition of `in` here allows the following code to compile. That makes the user able to pass a more specific version of a delegate without having to wrap it.:

```csharp
class Superclass { public int X; }
class Subclass : Superclass {}

...

FieldRef<Superclass, int> fsuper = x => ref x.X;

// old way
FieldRef<SubClass, int> fsub = x => ref fsuper(x);

// New way - this doesn't compile without `in`
FieldRef<Subclass, int> fsub = fsuper;
```

pardeike:
How does this correlate to scenarios where the field is private? FieldRef is mainly for private fields.

Therzok:
There's that icky case when some class has a `internal new int MyField;` that shadows the original field, so you need to do checks to see if the one you're getting is the right one.

By having this, you can use the same delegate type for both. Edge-case, but possibly useful.

pardeike:
At this point I am feeling confined enough to accept this PR but I fear that I still don't have the confidence to document this in a fitting way. Would you mind updating the api documentation too? I will pull it right away thereafter.

pardeike:
I use docfx so it's a matter of updating the /// comments and optionally the markdown files that are used to auto-generate the articles

Therzok:
Oh, oops, I completely forgot about the pull request, I am sorry!

Therzok:
I went and started writing a doc comment, and then I saw the line mentioning variance above.

Basically, the code change now makes the implementation behave as the documentation states. The variance for `T` was implicit, but not for delegate assignment.

Therzok:
I think no change is required?

PR #398, opened 2021-04-22
Bugfix priority attribute

Windows10CE:
When the HarmonyMethods for attributes are merged, Harmony considers that if the detail attribute field has a null value, it shouldn't be applied. This doesn't consider that the priority field defaults to -1, and causes it to be overridden if the HarmonyPriority attribute isn't merged last.

This is shown with 0b5867c32e23df54393a9d7ade907fe9da2f310b, which fails without the second commit also applied.

I think the proper solution would be to make priority a nullable int so it is caught by the existing if statements, but I *believe* this breaks API compat? Adding an exception to the ifs seemed like a safer option, so this PR does that instead.

### DIFF
--- a/Harmony/Public/HarmonyMethod.cs
+++ b/Harmony/Public/HarmonyMethod.cs
@@ -139,7 +139,10 @@ namespace HarmonyLib
 				HarmonyFields().ForEach(f =>
 				{
 					var val = trv.Field(f).GetValue();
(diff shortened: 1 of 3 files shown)


pardeike:
Ah, yes. Good catch. And also correct that a change would break the API so your PR is a good enough workaround.

PR #412, opened 2021-07-08
Fixes HarmonySharedState:GetOriginal  

simplyWiri:
Adds a unit test demonstrating, and then fixes a bug where the `MethodInfo` reference returned from a StackTrace is not equivalent to the reference which gets returned from `HarmonyLib.MethodPatcher:CreateReplacement`.

This relies on an update from MonoMod.Common which attempts to source the original method from method which is returned from the stack trace. This can then be used to find the pointer which represents the start of the method, which can be used to identify which original method it is referring to.

Let me know if you would like me to squash changes into a single commit, or clean up any details of the implementation.

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -1,7 +1,7 @@
 <Project>
 
 	<PropertyGroup>
-		<MonoModCommonVersion>21.6.21.1</MonoModCommonVersion>
+		<MonoModCommonVersion>21.7.13.2</MonoModCommonVersion>
 	</PropertyGroup>
 
 </Project>
--- a/Harmony/Internal/HarmonySharedState.cs
+++ b/Harmony/Internal/HarmonySharedState.cs
@@ -20,6 +20,9 @@ namespace HarmonyLib
 		internal const int internalVersion = 101;
 		internal static int actualVersion = -1;
 
+		// typeof(StackFrame).methodAddress
+		private static FieldInfo methodAddress = null;
+		
 		static T WithState<T>(Func<T> action)
 		{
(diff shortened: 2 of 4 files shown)


pardeike:
That PR is a backwards compatibility disaster waiting...

pardeike:
What’s wrong with my approach in < https://gist.github.com/pardeike/ba2d3837ca9e2938a0a1d34276495ef5> ?

simplyWiri:
PR needs some cleanup, but I think this will work now, from both a backwards compatibility & functionality standpoint. The things left to do
- [ ]  Move the added reflection out of Harmony.cs (probably something you might want to have a word on, as for where)
- [ ]  Get rid of the stuff which is failing on ubuntu/mac, which should fix the remaining errors
- [ ]  Remove all the nonsense spacing changes that accidentally got introduced
- [ ]  Cleanup the git history

pardeike:
I reran the test and it consistently fails on Win x86, macOS and Linux. Right now, I don't have the time to check in detail on what is happening. Are you still working on that for now?

pardeike:
One thing I wonder is: will the test you wrote fail if your code changes are not applied? If not, we don’t have a potent enough test.

simplyWiri:
It should, previously when I pushed the first test by itself, it would fail. I am not sure if the artefacts from that original commit still exist however (given I both rewrote the history, and I believe they are deleted after ~10 days).

pardeike:
Alright, I would like the test to be cleaned up a tad but I pull this for now and do it myself later. Good work, Wiri!

PR #421, opened 2021-09-09
Support for `JsonSerialization` where `BinaryFormatter` is not supported in .NET 5.0+

nwestfall:
Starting with I believe .NET 5, `BinaryFormatters` aren't supported by default.  To use them, you have to enable them in your .csproj with the following.
```xml
<EnableUnsafeBinaryFormatterSerialization>true</EnableUnsafeBinaryFormatterSerialization>
```

For security reasons, Microsoft is recommending that support for this be removed in applications as outlined here - https://docs.microsoft.com/en-us/dotnet/standard/serialization/binaryformatter-security-guide

The two places I can see this occurring (at least in the project I am using) are in the following methods
https://github.com/pardeike/Harmony/blob/4c6eaed10c3fd5133b8ee5805e1919103ca812ba/Harmony/Public/Patch.cs#L41-L47

https://github.com/pardeike/Harmony/blob/4c6eaed10c3fd5133b8ee5805e1919103ca812ba/Harmony/Public/Patch.cs#L53-L58

Using the recommendations mentioned in that article, I replaced the `BinaryFormatter` logic with `JsonSerializer` in `System.Text.Json`.  It does not serialize it to a JSON string, but rather still keeps it in bytes so that the rest of the logic works exactly as it did before.

To try and ensure backwards compatibility, I added compiler flags so that this change is only present in the following frameworks defined in the `.csproj`.
`netcoreapp3.0;netcoreapp3.1;net5.0`

Please let me know if you have any questions or want to discuss this PR more.  Thanks!

### DIFF
--- a/Harmony/Public/Patch.cs
+++ b/Harmony/Public/Patch.cs
@@ -5,6 +5,11 @@ using System.Reflection;
 using System.Reflection.Emit;
 using System.Runtime.Serialization;
 using System.Runtime.Serialization.Formatters.Binary;
+#if CORE_OR_STANDARD
+using System.Collections.Generic;
+using System.Text.Json;
+using System.Text.Json.Serialization;
+#endif
 
 namespace HarmonyLib
 {
@@ -34,16 +39,33 @@ namespace HarmonyLib
 			}
 		}
 
+#if CORE_OR_STANDARD
+		static JsonSerializerOptions _defaultJsonSerializerOptions = new JsonSerializerOptions()
+		{
+			IncludeFields = true,
+			Converters =
+			{
+				new PatchJsonConverter()
+			}
+		};
+#endif
+
 		/// <summary>Serializes a patch info</summary>
 		/// <param name="patchInfo">The <see cref="PatchInfo"/></param>
 		/// <returns>The serialized data</returns>
 		///
 		internal static byte[] Serialize(this PatchInfo patchInfo)
 		{
+			// ! https://docs.microsoft.com/en-us/dotnet/standard/serialization/binaryformatter-security-guide
+#if CORE_OR_STANDARD
+			var test = JsonSerializer.Serialize<PatchInfo>(patchInfo, _defaultJsonSerializerOptions);
+			return JsonSerializer.SerializeToUtf8Bytes<PatchInfo>(patchInfo, _defaultJsonSerializerOptions);
+#else
 			using var streamMemory = new MemoryStream();
 			var formatter = new BinaryFormatter();
 			formatter.Serialize(streamMemory, patchInfo);
 			return streamMemory.GetBuffer();
+#endif
 		}
 
 		/// <summary>Deserialize a patch info</summary>
@@ -52,9 +74,14 @@ namespace HarmonyLib
 		///
 		internal static PatchInfo Deserialize(byte[] bytes)
 		{
+			// ! https://docs.microsoft.com/en-us/dotnet/standard/serialization/binaryformatter-security-guide
+#if CORE_OR_STANDARD
+			return JsonSerializer.Deserialize<PatchInfo>(bytes, _defaultJsonSerializerOptions);
+#else
 			var formatter = new BinaryFormatter { Binder = new Binder() };
 			var streamMemory = new MemoryStream(bytes);
 			return (PatchInfo)formatter.Deserialize(streamMemory);
+#endif
 		}
 
 		/// <summary>Compare function to sort patch priorities</summary>
@@ -295,8 +322,8 @@ namespace HarmonyLib
 			set
 			{
 				patchMethod = value;
-				methodToken = patchMethod.MetadataToken;
-				moduleGUID = patchMethod.Module.ModuleVersionId.ToString();
+				methodToken = patchMethod?.MetadataToken ?? 0;
+				moduleGUID = patchMethod?.Module.ModuleVersionId.ToString();
 			}
 		}
 
@@ -328,6 +355,25 @@ namespace HarmonyLib
 		/// <param name="owner">An owner (Harmony ID)</param>
 		public Patch(HarmonyMethod method, int index, string owner)
 			: this(method.method, index, owner, method.priority, method.before, method.after, method.debug ?? false) { }
+
+		/// <summary>Creates a patch</summary>
+		/// <param name="patch">The method of the patch</param>
+		/// <param name="index">Zero-based index</param>
+		/// <param name="owner">An owner (Harmony ID)</param>
+		/// <param name="priority">The priority, see <see cref="Priority"/></param>
+		/// <param name="before">A list of Harmony IDs for patches that should run after this patch</param>
+		/// <param name="after">A list of Harmony IDs for patches that should run before this patch</param>
+		/// <param name="debug">A flag that will log the replacement method via <see cref="FileLog"/> every time this patch is used to build the replacement, even in the future</param>
+		/// <param name="methodToken">A token for method used for <see cref="PatchMethod"/></param>
+		/// <param name="moduleGUID">A GUID for the module used for <see cref="PatchMethod"/></param>
+		///
+		internal Patch(MethodInfo patch, int index, string owner, int priority, string[] before, string[] after, bool debug, int methodToken, string moduleGUID)
+			: this(patch, index, owner, priority, before, after, debug)
+		{
+			this.methodToken = methodToken;
+			this.moduleGUID = moduleGUID;
+		}
+
 
 		/// <summary>Get the patch method or a DynamicMethod if original patch method is a patch factory</summary>
 		/// <param name="original">The original method/constructor</param>
@@ -372,4 +418,84 @@ namespace HarmonyLib
 			return PatchMethod.GetHashCode();
 		}
 	}
+
+#if CORE_OR_STANDARD
+	class PatchJsonConverter : JsonConverter<Patch>
+	{
+		public override Patch Read(
+				ref Utf8JsonReader reader,
+				Type typeToConvert,
+				JsonSerializerOptions options)
+		{
+			if(reader.TokenType != JsonTokenType.StartObject)
+			{
+				throw new JsonException();
+			}
+
+			reader.Read(); //index
+			reader.Read();
+			int index = reader.GetInt32();
+			reader.Read(); //debug
+			reader.Read();
+			bool debug = reader.GetBoolean();
+			reader.Read(); //owner
+			reader.Read();
+			string owner = reader.GetString();
+			reader.Read(); //priority
+			reader.Read();
+			int priority = reader.GetInt32();
+			reader.Read(); //methodToken
+			reader.Read();
+			int methodToken = reader.GetInt32();
+			reader.Read(); //moduleGUID
+			reader.Read();
+			string moduleGUID = reader.GetString();
+			reader.Read(); // after
+			reader.Read();
+			List<string> after = new List<string>();
+			while(reader.Read())
+			{
+				if(reader.TokenType == JsonTokenType.EndArray)
+					break;
+				after.Add(reader.GetString());
+			}
+			reader.Read(); // before
+			reader.Read();
+			List<string> before = new List<string>();
+			while(reader.Read())
+			{
+				if(reader.TokenType == JsonTokenType.EndArray)
+					break;
+				before.Add(reader.GetString());
+			}
+			reader.Read();
+
+			return new Patch(null, index, owner, priority, before.ToArray(), after.ToArray(), debug, methodToken, moduleGUID);
+		}
+
+		public override void Write(
+			 Utf8JsonWriter writer,
+			 Patch patchValue,
+			 JsonSerializerOptions options)
+		{
+
+			writer.WriteStartObject();
+			writer.WriteNumber("index", patchValue.index);
+			writer.WriteBoolean("debug", patchValue.debug);
+			writer.WriteString("owner", patchValue.owner);
+			writer.WriteNumber("priority", patchValue.priority);
+			writer.WriteNumber("methodToken", patchValue.PatchMethod.MetadataToken);
+			writer.WriteString("moduleGUID", patchValue.PatchMethod.Module.ModuleVersionId.ToString());
+			writer.WriteStartArray("after");
+			foreach(var a in patchValue.after)
+				writer.WriteStringValue(a);
+			writer.WriteEndArray();
+			writer.WriteStartArray("before");
+			foreach(var b in patchValue.before)
+				writer.WriteStringValue(b);
+			writer.WriteEndArray();
+			writer.WriteEndObject();
+		}
+	}
+#endif
 }
--- a/Harmony/Public/Patch.cs
+++ b/Harmony/Public/Patch.cs
@@ -18,14 +18,22 @@ namespace HarmonyLib
 	internal static class PatchInfoSerialization
 	{
 #if NET50_OR_GREATER
+		static bool? _useBinaryFormatter = null;
 		static bool useBinaryFormatter
 		{
 			get
 			{
-				// https://github.com/dotnet/runtime/blob/208e377a5329ad6eb1db5e5fb9d4590fa50beadd/src/libraries/System.Runtime.Serialization.Formatters/src/System/Runtime/Serialization/LocalAppContextSwitches.cs#L14
-				bool hasSwitch = AppContext.TryGetSwitch("System.Runtime.Serialization.EnableUnsafeBinaryFormatterSerialization", out bool isEnabled);
-				// Default true - https://github.com/dotnet/runtime/blob/208e377a5329ad6eb1db5e5fb9d4590fa50beadd/src/libraries/Common/src/System/LocalAppContextSwitches.Common.cs#L54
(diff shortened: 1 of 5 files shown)


Harmony/Public/Patch.cs
```diff
@@ -295,8 +321,8 @@ public MethodInfo PatchMethod
 			set
 			{
 				patchMethod = value;
-				methodToken = patchMethod.MetadataToken;
-				moduleGUID = patchMethod.Module.ModuleVersionId.ToString();
```
pardeike, line 299: Can you elaborate on these changes?

nwestfall: Yes!  So I can't serialize "PatchMethod", so I instead pass a `null` value in the constructor.

https://github.com/nwestfall/Harmony/blob/e30e0a1452b8cd5ce8e1e07963a0b0294416b869/Harmony/Public/Patch.cs#L472

You get `NullReferenceExceptions` then when it tries to update `methodToken` and `moduleGUID`.

If you don't want me to edit the variable, I can just `null` check in the constructor.

pardeike: I understand the implications and the effect but not the cause. Why wouldn’t you be able to serialize a patch method?

nwestfall: This is because they contain `System.Type` fields, which cannot be serialized.

If you do, you get the following exception
```
Unhandled exception. System.NotSupportedException: Serialization and deserialization of 'System.Type' instances are not supported and should be avoided since they can lead to security issues. Path: $.DeclaringType.
```

Example code: https://dotnetfiddle.net/xszwaE

nwestfall: I did check and the `PatchMethod` gets "rebuilt" after the next `get` since there is a `null` check above it.  It only gets rebuilt though, if I set the `methodToken` and `moduleGUID`, which I can serialize.

pardeike: The original code never serialized the MethodInfo because in some .NET implementations you cannot serialize MethodBase or Type. Instead it only serializes the two aggregate fields (see 274 for the do-not-serialize annotation and 276+277 for the substitution fields.

I think you need to make your serialization do the same as it did before.

nwestfall: So I don't serialize `patchMethod` or `PatchMethod` either.  I serialize `methodToken` (276) and `moduleGUID` (277).  I accomplish this by writing my own serializer for `Patch` and do it in the following lines.

https://github.com/pardeike/Harmony/blob/e30e0a1452b8cd5ce8e1e07963a0b0294416b869/Harmony/Public/Patch.cs#L486-L487

When I then read the serialization, I pass them into a new constructor.
https://github.com/pardeike/Harmony/blob/e30e0a1452b8cd5ce8e1e07963a0b0294416b869/Harmony/Public/Patch.cs#L472

`PatchMethod`/`patchMethod` is never serialized and I rely on the same two fields you do `276` and `277`.

I think the confusion is I'm setting `PatchMethod` as `null` cause it's in the constructor.  If you want to leave it as it, I can make a new constructor where I don't need to pass `PatchMethod` so that way I don't need to `null` check.

nwestfall: This is the constructor I would update to not have `patch` and then not call the other constructor, but rather copy the logic.

https://github.com/pardeike/Harmony/blob/e30e0a1452b8cd5ce8e1e07963a0b0294416b869/Harmony/Public/Patch.cs#L369-L374

Would turn into

```c#
internal Patch(int index, string owner, int priority, string[] before, string[] after, bool debug, int methodToken, string moduleGUID)
{
        this.index = index;
	this.owner = owner;
	this.priority = priority == -1 ? Priority.Normal : priority;
	this.before = before ?? new string[0];
	this.after = after ?? new string[0];
	this.debug = debug;
	this.methodToken = methodToken;
	this.moduleGUID = moduleGUID;
}
```

pardeike:
So after I loaded in your PR as a new branch into Visual Studio and studying it, I think I would like to change so much about it that it would take quite a while to discuss here back and forth. I am inconclusive how we approach this. At one hand, I don't want to deny you the PR but on the other hand, I think it's not well written and misses a lot of points. Just going through the serializer documentation shows me that it actually serializes all getter which is probably not what you want. 

Plus your PR does not follow the formatting standard and it does not exclude the parts that are no longer needed when conditionally compiling code etc.

So how do we do this. Suggestions?

nwestfall:
> So after I loaded in your PR as a new branch into Visual Studio and studying it, I think I would like to change so much about it that it would take quite a while to discuss here back and forth. I am inconclusive how we approach this. At one hand, I don't want to deny you the PR but on the other hand, I think it's not well written and misses a lot of points. Just going through the serializer documentation shows me that it actually serializes all getter which is probably not what you want.
> 
> Plus your PR does not follow the formatting standard and it does not exclude the parts that are no longer needed when conditionally compiling code etc.
> 
> So how do we do this. Suggestions?

I'm happy to help and see this change through however I can.  I would be happy to jump on a phone call (zoom/teams/whatever) to maybe have you explain to me what might be the better approach.

From my perspective, this solution at a high level does work.  The tests past, I tried to keep .NET Framework the same with no breaking changes and the sample (and production) version of an application that specifically broken on .NET 5.0 works with these edits.  I can walk you through that too, maybe to show what and why I was thinking about this change.

I apologize about the formatting, I'm using VS Code on a Mac, so it probably won't clean up like Visual Studio would.  I really appreciate this library, the benefit and all your work to this point just on my PR alone.  If you are willing to meet, I'm happy to see if we can come up with a path forward.  If you don't want to meet, but maybe do this over email/slack, I can provide that too.

pardeike:
Join the discord and ping me. We can discuss this in #harmony-contributors and if you want also via audio. I’m on Sweden time.

Umaaz:
Hi, this issue is a problem for us. We use harmony in our product and we need to support .NET 3 and 5. I am not the most experienced .NET dev. Looking through this change it seems to make sense.

If you need any assistance in implementing a change, or testing etc let me know. 😃

pardeike:
I am planning to work with @nwestfall to make this PR happen. In its current form there are few things to iron out first.

nwestfall:
I want to do some more testing, but I took a pass at most of what we discussed today.  Few things while implementing I wanted to call out, to see if it warrants further discussion and maybe some further improvement.

For `BinaryFormatter`, there is no way to check if it's enabled in .NET 5.0 other than running it catching a `NotSupportedException`.

```c#
		internal static byte[] Serialize(this PatchInfo patchInfo)
		{
#if NET50_OR_GREATER
			try
			{
#endif
			using var streamMemory = new MemoryStream();
			var formatter = new BinaryFormatter();
			formatter.Serialize(streamMemory, patchInfo);
			return streamMemory.GetBuffer();
#if NET50_OR_GREATER
			}
			catch(NotSupportedException) // ! https://docs.microsoft.com/en-us/dotnet/standard/serialization/binaryformatter-security-guide
			{
				return JsonSerializer.SerializeToUtf8Bytes<PatchInfo>(patchInfo);
			}
#endif
		}
```
https://github.com/nwestfall/Harmony/blob/d82d93715a648eaa3e17aecc216a8fa0768ae170/Harmony/Public/Patch.cs#L45-L62

Question I have, is that to improve this performance, should I flag after the first time the exception is thrown so that I don't always need to do the `try/catch` block?  If it fails with the `NotSupportException` at least once, I know that it is off and can just start always using JSON.  I personally would, but I'm curious your thoughts here.

I also cleaned up the `JsonConverter` like we discussed.  I was able to completely remove the `JsonSerializationOptions` block in favor of the attributes.  This should hopefully better match what you currently serialize and clean it up a bit.  I also moved the `PatchJsonConverter` to a new file, flagged for .NET 5.0+ and created some extensions to better explain it.  To only compile these files for .NET 5.0+, I added the following group to the `.csproj`

```xml
<ItemGroup Condition="!$(IsNET5OrGreater)">
   <Compile Remove="**\*.net5.cs" />
</ItemGroup>
```

I remember now why I had to call `Read()` twice.  This is because you have to `read` the property name, then the value.  Since we don't care about the property name, we call it twice in a row.  This helper should help make it easier to read.

```c#
internal static void SkipPropertyName(this Utf8JsonReader reader)
{
   reader.Read(); //this is the `PropertyName`, skip it
   reader.Read(); //this is the `value`, making it ready to get
}
```

I still have some work and want to work on testing that edge case you mentioned with the two different versions, but I wanted to share progress thus far to see if you had any further feedback on the items.

pardeike:
I would create a method that instantiates the binary formatter inside a try catch and that returns a book. Then call that from a static initializer and set a binary available flag that you then use later. That is a pattern we use in Memory too so you can match the style. Maybe even add the method there instead of in Patch.cs since it’s about the platform.

Umaaz:
From the code of the binary formatter https://github.com/dotnet/runtime/blob/57bfe474518ab5b7cfe6bf7424a79ce3af9d6657/src/libraries/System.Runtime.Serialization.Formatters/src/System/Runtime/Serialization/Formatters/Binary/BinaryFormatter.Core.cs

It seems that the exception is thrown on serialize not from the constructor?

Could also be worth looking at how they check for the app flag which I think is here: https://github.com/dotnet/runtime/blob/57bfe474518ab5b7cfe6bf7424a79ce3af9d6657/src/libraries/System.Runtime.Serialization.Formatters/src/System/Runtime/Serialization/LocalAppContextSwitches.cs#L8

nwestfall:
Looks like what @Umaaz found with `AppContext` can be done!  I'm going to run some more tests, but I added this check in favor of the `try/catch`.  If the switch cannot be found, it will default to `true`, which matches Microsoft

https://github.com/dotnet/runtime/blob/208e377a5329ad6eb1db5e5fb9d4590fa50beadd/src/libraries/Common/src/System/LocalAppContextSwitches.Common.cs#L54

```c#
#if NET50_OR_GREATER
static bool useBinaryFormatter
   {
      get
      {
         bool hasSwitch = AppContext.TryGetSwitch("System.Runtime.Serialization.EnableUnsafeBinaryFormatterSerialization", out bool isEnabled);
         return (hasSwitch) ? isEnabled : true;
      }
}
#endif
```

I'm going to run some more tests, just to be sure before I check it in.

Also @pardeike , let me know if there is a place you want me to log this flag.  I see cases of things being logged in `FileLog.Log` if `DEBUG` is on and this might be a good one to add.

pardeike:
I think we don’t need to log the rather internal behavior of the serializer. It pollutes the log more than it helps. The value can be deduced from the environment and Harmony itself does not log much because it could be running in situations where you simply cannot log. The FileLog is mainly for debugging patching and does not contain normal status information.

nwestfall:
I think (hope) this PR is in another state for a review.  I've updated the following

 - `PatchInfo` now has `annotations` for serialization
 - `PatchInfoSerialization` has a new flag `useBinaryFormatter` for .NET 5.0+ used when serializing/deserializing
 - New `internal` constructor in `Patch` repeats setting fields, so that `PatchMethod` isn't changed
 - `PatchJsonConverter` is now in it's own file that only compiles with .NET 5.0
 - Created `Utf8JsonReaderExtensions` that compiles with .NET 5.0 to help with readability in `PatchJsonConverter`

Harmony/Public/Patch.cs
```diff
@@ -5,13 +5,31 @@
...
 	{
+#if NET50_OR_GREATER
+		static bool useBinaryFormatter
+		{
+			get
+			{
+				// https://github.com/dotnet/runtime/blob/208e377a5329ad6eb1db5e5fb9d4590fa50beadd/src/libraries/System.Runtime.Serialization.Formatters/src/System/Runtime/Serialization/LocalAppContextSwitches.cs#L14
+				bool hasSwitch = AppContext.TryGetSwitch("System.Runtime.Serialization.EnableUnsafeBinaryFormatterSerialization", out bool isEnabled);
+				// Default true - https://github.com/dotnet/runtime/blob/208e377a5329ad6eb1db5e5fb9d4590fa50beadd/src/libraries/Common/src/System/LocalAppContextSwitches.Common.cs#L54
+				return (hasSwitch) ? isEnabled : true;
```
pardeike, line 28: I would love to have a
 ```cs
 if (hasSwitch) return isEnabled;
 return true; // explain why true here
 ```

Harmony/Public/Patch.cs
```diff
@@ -5,13 +5,31 @@
...
+	///
 	internal static class PatchInfoSerialization
 	{
+#if NET50_OR_GREATER
+		static bool useBinaryFormatter
+		{
+			get
+			{
+				// https://github.com/dotnet/runtime/blob/208e377a5329ad6eb1db5e5fb9d4590fa50beadd/src/libraries/System.Runtime.Serialization.Formatters/src/System/Runtime/Serialization/LocalAppContextSwitches.cs#L14
+				bool hasSwitch = AppContext.TryGetSwitch("System.Runtime.Serialization.EnableUnsafeBinaryFormatterSerialization", out bool isEnabled);
```
pardeike, line 26: Can this be expected to change during runtime or can we put it into a static bool?

Harmony/Public/PatchJsonConverter.net5.cs
```diff
@@ -0,0 +1,57 @@
+using System;
+using System.Collections.Generic;
+using System.Text.Json;
+using System.Text.Json.Serialization;
+
+namespace HarmonyLib
+{
+	internal class PatchJsonConverter : JsonConverter<Patch>
+	{
+		public override Patch Read(
```
pardeike, line 10: While many people disagree, I actually like long lines and they are used all over the project.

Harmony/Public/PatchJsonConverter.net5.cs
```diff
@@ -0,0 +1,57 @@
...
+				ref Utf8JsonReader reader,
+				Type typeToConvert,
+				JsonSerializerOptions options)
+		{
+			if(reader.TokenType != JsonTokenType.StartObject)
+			{
```
pardeike, line 16: Unnecessary {} can be removed as it is standard in the project

Harmony/Public/UtfJsonReaderExtensions.net5.cs
```diff
@@ -0,0 +1,34 @@
+using System.Collections.Generic;
+using System.Text.Json;
+
+namespace HarmonyLib
+{
+	internal static class Utf8JsonReaderExtensions
+	{
+		internal static void SkipPropertyName(this Utf8JsonReader reader)
+		{
+			reader.Read(); //this is the `PropertyName`, skip it
```
pardeike, line 10: sorry to be nitpicky but space after //

Harmony/Public/UtfJsonReaderExtensions.net5.cs
```diff
@@ -0,0 +1,34 @@
...
+			reader.Read(); //this is the `value`, making it ready to get
+		}
+
+		internal static void WriteStringArray(this Utf8JsonWriter writer, string propertyName, IEnumerable<string> array)
```
pardeike, line 14: Can you find a better name than `array`?

Harmony/Public/UtfJsonReaderExtensions.net5.cs
```diff
@@ -0,0 +1,34 @@
...
+		{
+			writer.WriteStartArray(propertyName);
+			foreach(var a in array)
```
pardeike, line 17: Same here, probably go with `str in strings`

Harmony/Public/UtfJsonReaderExtensions.net5.cs
```diff
@@ -0,0 +1,34 @@
...
+				writer.WriteStringValue(a);
+			writer.WriteEndArray();
+		}
+
+		internal static List<string> GetStringArray(this Utf8JsonReader reader)
+		{
+			List<string> data = new List<string>();
```
pardeike, line 24: `array` -> `strings`

Harmony/Public/UtfJsonReaderExtensions.net5.cs
```diff
@@ -0,0 +1,34 @@
...
+			while(reader.Read())
+			{
+				if(reader.TokenType == JsonTokenType.EndArray)
```
pardeike, line 27: Move negated into the while condition maybe:
`while(reader.read() && reader.TokenType != JsonTokenType.EndArray)`

pardeike:
With the small changes I suggested added, I can accept this.

Harmony/Public/UtfJsonReaderExtensions.net5.cs
nwestfall: I can maybe do one of the following?  I was trying to keep it inline with the existing `WriteStartArray` function in `System.Text.Json`

- [ ] `WriteStringData`
- [ ] `WriteStrings`
- [ ] `WriteStringList`
- [ ] `WriteStringSet`

Harmony/Public/Patch.cs
nwestfall: It looks like .NET caches it, so we can do the same

Umaaz:
Great to see this accepted! Any chance we can get a release with these changes?

pardeike:
After testing all changes that are on master since last release, yes. If you are impatient you can build master yourself.

PR #432, opened 2021-10-17
Allowing TargetMethod(s) with ReversePatch

kohanis:
Now ReversePatch should pick target from TargetMethod (and PatchAll  or TargetMethods if only 1 method) if not annotated

### DIFF
--- a/HarmonyTests/ReversePatching/Assets/ReversePatches.cs
+++ b/HarmonyTests/ReversePatching/Assets/ReversePatches.cs
@@ -91,4 +91,25 @@ namespace HarmonyLibTests.Assets
 			return false;
 		}
 	}
+
+	[HarmonyPatch]
+	public class Class1ReversePatchWithTargetMethod
+	{
+		public static MethodBase TargetMethod()
+		{
+			return AccessTools.Method(typeof(Class1Reverse), "GetExtra");
+		}
+
+		[HarmonyReversePatch]
+		[MethodImpl(MethodImplOptions.NoInlining)]
+		public static string GetExtra(int n)
+		{
+			// this will be replaced by reverse patching
+
+			// using a fake while loop to force non-inlining
+			while (DateTime.Now.Ticks > 0)
+				throw new NotImplementedException();
+			return null;
+		}
+	}
 }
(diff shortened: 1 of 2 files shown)


pardeike:
Interesting. Could you please add at least one test to verify that this works as intended?

kohanis:
Done. Fits?

kohanis:
Actually, after thinking about it, now I see that it can break code in edge cases where reverse patch + transpiler used to define TargetMethod(s) result(s). While there is probably no such code, it would be a good idea to point out in docs that auxilary methods executed before reverse patches.

pardeike:
Can you give me a minimal example case in pseudo code that would not work as expected? I assume it’s when you have a patch class with a TargetMethod() and a reverse patch and for example a prefix together.

kohanis:
Not what I mean. For example, as my situation: 
```csharp
using System;
using System.Collections.Generic;
using System.Linq;
using System.Reflection;
using System.Reflection.Emit;
using HarmonyLib;

class Original {
    void fn() {
        // yeah, it just like this for some reasons
        fn2(delegate {
            // some code I need to reach
            Console.WriteLine("Hello there");
        });
    }
    
    void fn2(Action action)
    {
        action();
    }
}

[HarmonyPatch(typeof(Original), "fn")]
class Original_Patch {
    static void fn_delegate()
    {
        IEnumerable<CodeInstruction> Transpiler(IEnumerable<CodeInstruction> instructions) {
            var list = instructions.ToList();
            int op = list.FindIndex(item => item.opcode == OpCodes.Ldftn);
            Original_delegate_Patch.TargetMethodInfo = list[op].operand as MethodInfo;
            
            yield return list[op - 1];
            yield return new CodeInstruction(OpCodes.Call, list[op].operand);
            yield return new CodeInstruction(OpCodes.Ret);
        }
            
        _ = Transpiler(null);
    }
}

class Original_delegate_Patch {
    internal static MethodInfo TargetMethodInfo;

    private static MethodBase TargetMethod() {
        return TargetMethodInfo;
    }
    
    public static void Postfix() {
        // some code
    }
}
```
Will not work after this change. But, as I said, this is a hack to not iterate method body twice. Still can be done with CreateReversePatcher(...).Patch() manually

kohanis:
Oh, okay, I just checked PatchAll, we just should not interact directly between patches like this with PatchAll, I get it

pardeike:
Which means that this does not work as intended? Should I roll back the PR? Sorry but right now, I am busy with something else and can’t check all your changes in detail.

kohanis:
No, everything should be fine

PR #437, opened 2021-12-02
Improve Performance of `FindReplacement`

simplyWiri:
I was experimenting with some code which used FindReplacement very frequently (to analyze a large number of heavily modded stack-traces). I noticed that a non trivial amount of time was spent in this method, when it didn't have to.

The two main existing issues are:
1. The actual lookup `.GetNativeStart()` is relatively expensive, and in the average case, is executed on half of the methods that harmony stores, every time someone calls `FindReplacement`.
2. The code, while doing this expensive lookup, was inside a lock used when patching/unpatching etc. 

This is one approach to fixing both issues, solving it by caching the native start of methods inside HarmonySharedState. The locking is subsequently no longer an issue, because the lookup is now significantly faster. 

I also believe this approach is backwards compatible, because we can re-calculate this field from the values inside `originals`.

### DIFF
--- a/Harmony/Internal/HarmonySharedState.cs
+++ b/Harmony/Internal/HarmonySharedState.cs
@@ -33,11 +33,13 @@ namespace HarmonyLib
 	internal static class HarmonySharedState
 	{
 		const string name = "HarmonySharedState";
-		internal const int internalVersion = 101; // bumb this if the layout of the HarmonySharedState type changes
+		internal const int internalVersion = 102; // bump this if the layout of the HarmonySharedState type changes
 
-		// state/originals are set to instances stored in the global dynamic types static fields with the same name
+		// state/originals/methodStarts are set to instances stored in the global dynamic types static fields with the same name
 		static readonly Dictionary<MethodBase, byte[]> state;
 		static readonly Dictionary<MethodInfo, MethodBase> originals;
+		// maps the native start of the method to the method itself
+		static readonly Dictionary<long, MethodInfo> methodStarts;
 		internal static readonly int actualVersion;
 
 		static HarmonySharedState()
@@ -68,6 +70,12 @@ namespace HarmonyLib
 			originals = new Dictionary<MethodInfo, MethodBase>();
 			if (originalsField != null) // may not exist in older versions
 				originals = (Dictionary<MethodInfo, MethodBase>)originalsField.GetValue(null);
+
+			// re-create 'methodStarts' based on the value(s) in 'originals'
+			methodStarts = new Dictionary<long, MethodInfo>();
+			foreach ( var original in originals.Keys ) {
+				methodStarts.Add(original.GetNativeStart().ToInt64(), original);
+			}
 
 			// newer .NET versions can re-jit methods so we need to patch them after that happens
 			DetourHelper.Runtime.OnMethodCompiled += (MethodBase method, IntPtr codeStart, ulong codeLen) =>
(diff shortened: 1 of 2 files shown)


0x0ade:
> The actual lookup `.GetNativeStart()` is relatively expensive, and in the average case, is executed on half of the methods that harmony stores, every time someone calls `FindReplacement`.

It is expensive and uncached because there are no promises made for the start of a method to stay the same for as long as the process is running.
- .NET Core 3.1+ introduces re-JITing. One *could* cache the native start of a method and invalidate / refresh the cache when it gets re-JITed, but...
- Even besides re-JITing, there could always be something else around the corner that messes with the method's starting point. Case in point, it was recently discovered that Wine adds yet another indirection for AOT-compiled methods, presumably as part of its .dll replacement mechanisms. This led to actual problems in at least one community.

If absolutely necessary, please make sure that the cache takes re-JITing into account and can be toggled via f.e. an environment flag.

simplyWiri:
> If absolutely necessary, please make sure that the cache takes re-JITing into account and can be toggled via f.e. an environment flag.

Should I be looking at extending the code here to insert a call to implement that functionality? [HarmonySharedState](https://github.com/pardeike/Harmony/blob/master/Harmony/Internal/HarmonySharedState.cs#L72-L79)
```csharp
			// newer .NET versions can re-jit methods so we need to patch them after that happens
			DetourHelper.Runtime.OnMethodCompiled += (MethodBase method, IntPtr codeStart, ulong codeLen) =>
			{
				if (method == null) return;
				var info = GetPatchInfo(method);
				if (info == null) return;
				PatchFunctions.UpdateRecompiledMethod(method, codeStart, info);
			};
```
And, also, would there be a way to find the old codeStart from the method which has just been recompiled? (otherwise the caching will just 'leak' memory each time a new entry is added), assuming that calling `.GetNativeStart()` on the new `MethodBase` will return a value equal to `codeStart`, not necessarily to the old value which we have cached (as the key). Also worth mentioning that MethodInfo comparisons is why I switched to using the native start of a method, so trying to remove entries based on their implementation of `==` for MethodInfo would not be consistent either.

As a tldr, options that I could see working;
- Updating the cache when a method gets recompiled ( in the OnMethodCompiled lambda, add a check to see if `info` is not null, and `codeStart` does not exist in the `methodStarts`dictionary, if it does not, add it. ('leaks' memory for the just-replaced entry)
- The same as the last one, but instead of just updating one method, refresh the entire cache using the values in `originals`. Would be considerably slower, but wouldn't leak memory.

> Even besides re-JITing, there could always be something else around the corner that messes with the method's starting point.

As long as its consistent, it doesn't matter if the `methodStart` is actually pointing to the first instruction, or some trampoline instruction(s). This is primarily used in recovery of original patch methods from replacements in stack traces.

0x0ade:
> Should I be looking at extending the code here to insert a call to implement that functionality?

That seems like the right place. Maybe inside of `UpdateRecompiledMethod` would be a better fit, given that this only concerns already patched methods? I feel like @pardeike will have a better idea on where to squeeze that cache invalidation in 😅

> And, also, would there be a way to find the old codeStart from the method which has just been recompiled?

I don't think so. MonoMod uses method reference comparisons under the hood, but it doesn't need to work around a shared state, thus I don't know what the best course of action would be for Harmony. My tired brain (will go to bed soon) does see a third option though: Maintain both a Dict<MethodBase, IntPtr> and a Dict<IntPtr, MethodBase>. When updating, get the already cached method start from the first dict first, and remove it from the second dict, then proceed with the usual cache update.

pardeike:
If you in OnMethodCompiled can invalidate that methods cache so it recaches when someone asks for the method start again, that would be a good way to tackle that.

pardeike:
Without more work done, I would like to retire this PR since its a relatively specific fix for a specific use case that is not very common.

PR #446, opened 2022-01-06
Add link to the Harmony deno discord library in case someone finds the wrong project

MarcelskyXD:


pardeike:
Sorry, I am not adding some link to a random project that has a similar name. Especially when it was started after this project. IMHO, adding the link just adds to the confusion.

PR #455, opened 2022-02-25
Removed unneccesary '.ToList().AsEnumerable()' in Harmony/Internal/CodeTranspiler.cs

ForsakenNydoseth:
The output of the '_Select_' method is already of type 'IEnumerable<CodeInstruction>' so enumerating it and then converting back to 'IEnumerable<CodeInstruction>' is **obsolete**.


pardeike:
Not sure which one you mean. Usually .ToList() has the effect of forcing the enumeration to immediately run and leaving it out can lead to unwanted delays in execution.

PR #465, opened 2022-03-22
Fix Unpatch default parameter

Windows10CE:
Unpatch eventually calls into Patch.Remove, where the logic checks for `"*"` as the "remove all" wildcard, not null, which is the current default of Unpatch.

### DIFF
--- a/Harmony/Public/Harmony.cs
+++ b/Harmony/Public/Harmony.cs
@@ -174,7 +174,7 @@ namespace HarmonyLib
 		/// <param name="type">The <see cref="HarmonyPatchType"/></param>
 		/// <param name="harmonyID">The optional Harmony ID to restrict unpatching to a specific Harmony instance</param>
 		///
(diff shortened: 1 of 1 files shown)


pardeike:
Is this a breaking change (like discussed in http://www.thebillwagner.com/blog/Item/2010-08-23-DegreesofBreakingChangesOptionalParametersinC40) ?

Windows10CE:
That article would count this change as a breaking one, yes, because the default behavior will change from doing nothing to unpatching all patches on a method. As I wouldn't expect a default optional parameter to have a method do nothing, the new default is likely what callers expected to happen.

So, there are two scenarios after a recompile:
1. Callers now get the expected behavior and everything works fine (not changing this here would've required a bugfix at the caller level)
2. Callers get expected behavior, but their code breaks as a result. However, this happening is basically a "I was relying on that broken code!".

Whether or not you consider it a breaking change depends on if you consider a bugfix that changes something to the expected behavior when people could've been relying (on accident) on the broken behavior to be a breaking change

PR #508, opened 2022-12-01
Adding .NET Standard 2.1 support

warkenji:


pardeike:
Support for .Net Standard was removed because System.Reflection.Emit was removed from Nuget and without it Harmony cannot do its job.

warkenji:
> System.Reflection.Emit

.Net Standard 2.1 don't need it :)

sunnamed434:
Any thoughts on adding it?

pardeike:
Thoughts? More than the existing statements made in this thread, no nothing has changed since my last comment.

PR #514, opened 2023-03-02
Adds support for targeting an async method's internal MoveNext method

Banane9:
As the title says. This is particularly interesting for targeting an `async` method with a transpiler, whether using Attributes or manually finding it with the added `AsyncMoveNext` method in AccessTools.

I'm not sure where the tests should go / how they should look, but I did test it on a game that I write mods for.  
Works perfectly, although trying to use `IAsyncStateMachine` as the type for the `__instance` parameter causes a `FormatException` for invalid IL on the call to the patch method (that took a while to figure out).

Definition of Patches

Resulting output in log showing all patches are successfully applied

### DIFF
--- a/Harmony/Public/Attributes.cs
+++ b/Harmony/Public/Attributes.cs
@@ -17,10 +17,10 @@ namespace HarmonyLib
 		Constructor,
 		/// <summary>This is a static constructor</summary>
 		StaticConstructor,
-		/// <summary>This targets the MoveNext method of the enumerator result</summary>
+		/// <summary>This targets the MoveNext method of the enumerator result, that actually contains the method's implementation</summary>
 		Enumerator,
-#if NET40_OR_GREATER
-		/// <summary>This targets the MoveNext method of the async state machine</summary>
+#if NET45_OR_GREATER
+		/// <summary>This targets the MoveNext method of the async state machine, that actually contains the method's implementation</summary>
 		Async
 #endif
 	}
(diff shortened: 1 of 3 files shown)


pardeike:
What’s the difference to [MethodType.Enumerator](https://harmony.pardeike.net/api/HarmonyLib.MethodType.html) ?

Banane9:
Well, with MethodType.Enumerator it fails to find the method to patch, as it looks for a newobj opcode - but the actual contents of async methods don't use that.

Looking at an iterator method, those are also decorated with an `IteratorStateMachineAttribute`, which contains a reference to the type. They both share the `StateMachineAttribute` as a parent, so I suppose it could be merged together into a StateMachine one too (though that wouldn't be backwards compatible, I guess).

Banane9:
Anything that would have to be changed/added for this? 👀

pardeike:
I would still like to see if we can merge this with the existing api. You mentioned that this would break backwards compatibility, in what way would that be a problem?

Banane9:
I just mentioned it as a consideration - I don't know how you handle that sort of thing for Harmony.

Funnily enough, the way I understand it, renaming the `MethodType.Enumerator` enum value to something like `MethodType.StateMachine` and handling both through that would keep it binary compatible (as enum values are embedded directly as their numeric values), but code using it would have to be changed.

Harmony/Public/Attributes.cs
```diff
@@ -18,7 +18,11 @@ public enum MethodType
 		/// <summary>This is a static constructor</summary>
 		StaticConstructor,
 		/// <summary>This targets the MoveNext method of the enumerator result</summary>
-		Enumerator
+		Enumerator,
+#if NET40_OR_GREATER
+		/// <summary>This targets the MoveNext method of the async state machine that actually has the method's implementation</summary>
```
pardeike, line 23: If you shorten the comment a bit (same style as AccessTools.cs:504) I'll approve the PR

pardeike (changes requested):
Just that one tiny cleanup

PR #519, opened 2023-04-03
Add targeting of indexer properties

Banane9:
This makes `MethodType`s `Getter` and `Setter` target indexer properties when used without a name.  
Type parameters are respected to target overloaded indexers.

Adds indexer specific versions of the Property methods to AccessTools, which take the containing Type and optional parameters.


pardeike:
Nice work!

PR #524, opened 2023-04-08
Added netstandard2.0 reference assemblies support (do not merge)

Aragas:


Aragas:
It has served it's purose

PR #534, opened 2023-06-27
Using `PlatformTriple.Current.GetIdentifiable` for a user submitted MethodInfo

Aragas:
This fixed some cases when Harmony couldn't return the original method for a patch method
Also updated the  MonoMod

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -3,7 +3,7 @@
 	<PropertyGroup>
 		<HarmonyVersion>2.3.0.0</HarmonyVersion>
 		<HarmonyPrerelease>-prerelease.3</HarmonyPrerelease>
-		<MonoModCoreVersion>1.0.0-prerelease.2</MonoModCoreVersion>
+		<MonoModCoreVersion>1.0.0</MonoModCoreVersion>
 	</PropertyGroup>
 
 </Project>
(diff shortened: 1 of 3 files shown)


pardeike:
I think we have a regression in the 1.0 version of MonoMod Core. I know Azure fails randomly but I got very consistent test failure in this PR due to the bump in MMC version. Ping @nike4613

Particularly, tests that have to do with exception block handling fail for net framework <= 4.8 and only on Linux&MacOS. 

Oh and thanks for the fix for stack item handling.

nike4613:
It looks like all the failures on Azure here are invalid IL failures-- are you certain that it's MMC that's causing them?

pardeike:
Pretty sure. I have successful tests on master that are recent: https://dev.azure.com/pardeike/Harmony/_build?definitionId=1

nike4613:
I believe [these](https://github.com/MonoMod/MonoMod/compare/8ede4737d89d0fb2cff4151830eec71662dd3a04..9444ac977a0a7af3fba3e375a4f3e4715ec31778) are the only actual changes from the prereleases

Aragas:
Should I remove the version update from the PR then?

pardeike:
Despite this failing three tests
- HarmonyLibTests.Patching.Arguments:Test_RefResults
- HarmonyLibTests.Patching.Specials:Test_PatchException
- HarmonyLibTests.Patching.CombinedPatches:Test_ManyFinalizers
that all have in common that they deal with original methods like:
```cs
public void Method1()
{
	throw new Exception();
}
```
I will pull this PR and solve the "why?" later. The problem is that mono does not like methods that don't end in a RET and throws an IL error. I can vagely remember that we had this problem before but how did we solve it? Cannot remember.

PR #537, opened 2023-07-11
GetExecutingAssembly Tuning and Tests

Aragas:
* We can utilities SymbolExtensions a bit better. Basically, we can add any input signature to the lambda, so we can add parameters that need `ref` and `out` parameters to be set

* I've added more "defensive" code to the frame resolution, just in case I suppose?

* I've added an additional test project to test that we are able to correctly get the executing assembly from "user code" basically. Checking it might be sufficient from the test project, but I would like to have a case when the entry assembly and the executing are not the same


TestLibrary/TestPatch.cs
```diff
@@ -0,0 +1,23 @@
...
+
+	public static void TestPatching(out Assembly original, out Assembly patched)
+	{
+		original = Assembly.GetExecutingAssembly();
+
+		var instance = new Harmony("test");
+		instance.Patch(SymbolExtensions.GetMethodInfo(() => Method()), transpiler: SymbolExtensions.GetMethodInfo(() => EmptyTranspiler(null)));
+
+		patched = Assembly.GetExecutingAssembly();
+	}
```
pardeike, line 22: Shouldn't we also test the case where the original method calls Assembly.GetExecutingAssembly?

Aragas: Can't believe I forgot that. Fixed in the latest commit

PR #556, opened 2023-10-21
Make AccessTools.all really give all members

Banane9:
Adds `BindingFlags.FlattenHierarchy` to `AccessTools.all` (and removes it on `declaredOnly`).

Otherwise static members defined in a parent class aren't returned.


pardeike:
After long consideration I think it’s too dangerous to change all. The benefits of it are low since the use case is rare and easily fixed by adding it yourself in that case. The problem with breaking existing usages that don’t expect a different result is too big.

PR #560, opened 2023-12-25
Fix typo in intro.md

jivanf:
Just a small typo in this sentence:
> Harmony uses a variation of hooking and **focuces** only on runtime changes that don't affect files on disk:

It should read like this:
> Harmony uses a variation of hooking and **focuses** only on runtime changes that don't affect files on disk:

Thank you for this library! Merry Christmas and happy New Year 🎊


pardeike:
Thanks and God Jul and Gott Nytt År to you too!

PR #578, opened 2024-02-13
Tests: Better mono setup on Windows

Aragas:


Aragas:
Not sure if we wanna merge.
So the mono download link I'm using (preview) in the script is not entirely correct - we need to use the non preview, but it returns a 2019 build. Preview seem to return the latest stable for now.
The choco kind guarantees the correct latest version, but it takes about 30 seconds more when installing mono.

PR #582, opened 2024-02-22
Way to override returned reference itself

kohanis:
It is, on the whole, a suggestion. Currently, there is no way to override the reference itself in a method like `ref T fn()`
This implementation adds an optional parameter `out RefResult<T> __resultRef` (`out` is a requirement as it is called immediately after the patch and nulled), where `delegate ref T RefResult<T>()`
I don't have a user-case for this, but someone might have one

// On second thought, `out` requirement is unnecessary, `ref` is okay

### DIFF
--- a/Harmony/Internal/MethodPatcher.cs
+++ b/Harmony/Internal/MethodPatcher.cs
@@ -15,6 +15,7 @@ namespace HarmonyLib
 		const string ORIGINAL_METHOD_PARAM = "__originalMethod";
 		const string ARGS_ARRAY_VAR = "__args";
 		const string RESULT_VAR = "__result";
+		const string RESULT_REF_VAR = "__resultRef";
 		const string STATE_VAR = "__state";
 		const string EXCEPTION_VAR = "__exception";
 		const string RUN_ORIGINAL_VAR = "__runOriginal";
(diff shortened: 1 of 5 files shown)


kohanis:
Lack of documentation is intentional until acceptance of feature itself is confirmed

HarmonyTests/Patching/Specials.cs
```diff
@@ -50,6 +50,34 @@ public void Test_HttpWebRequestGetResponse()
...
+			Assert.NotNull(instance);
+			var processor = instance.CreateClassProcessor(typeof(ResultRefStruct_Patch));
+			Assert.NotNull(processor, "processor");
+
+			test.Method1() = 1;
+			test.Method2() = 2;
+			test.Method3() = 3;
+			Assert.Throws<Exception>(() => test.Method4(), "Method4 does not throw");
+
+			Assert.AreEqual(new[] { 1, 0, 2, 0, 3, 0, 0 }, ResultRefStruct.Ints);
```
pardeike, line 68: Running test multiple times might fail it because the zeros are not zero anymore. Better add an initial reset of `ResultRefStruct.Ints`

Harmony/Internal/MethodPatcher.cs
```diff
@@ -571,7 +586,6 @@ void EmitCallParameter(MethodInfo patch, Dictionary<string, LocalBuilder> variab
 				// treat __result var special
 				if (patchParam.Name == RESULT_VAR)
 				{
-					var returnType = AccessTools.GetReturnedType(original);
```
pardeike, line 574: Weird. For me this looks like `returnType` isn't assigned anymore but used in the next line.

pardeike (changes requested):
Very nice work. I added two comments, please verify. And beside the comments, I think we can go and add some documentation (make sure you don't edit html files but appropiate markdown in /Harmony/Documentation and all docfx annotations of course.

Harmony/Internal/MethodPatcher.cs
kohanis: There's a field with same value in the class

kohanis:
Added docs(Never been good at it, hope it's understandable), new test + initial reset + restructured. Lifted `out` requirement

pardeike:
I will pull this and clean up any documentation if necessary later. Excellent work, much appreciated and great timing to get this into the next big release.

kohanis:
Actually, I was wrong. This could have been achieved through prefix with skip and/or passthrough postfix. I completely forgot about them. But now it's too late to delete... Apologies

PR #588, opened 2024-03-02
Added pdb generation

Aragas:
* for Thin they are embedded
* for Fat they are portable
Removed snupkg generation, the .pdb files are included in the nupkg

Side fix:
Exclude the secondary System.Reflection.* packages from NuGet - we were adding new dependencies that we don't really need


Aragas:
PR ready for merge

PR #598, opened 2024-03-07
GetMethodFromStackframe fix (and .gitignore update)

kohanis:


pardeike:
Why are the rather large changes to `.gitignore` necessary?

kohanis:
They are not, it's just latest template from dotnet cli. It at least adds files for JetBrains IDEs (I mostly use Rider for .net). I can remove it

pardeike:
So I played around with the change a bit but I am confused. What is the purpose of this change. Submitting a PR with no context usually only works for trivial changes but this areas is unexpectedly complicated due to how Mono works and the interactions around stacktraces. The risk of breaking things is high.

kohanis:
Fair
Currently there are:
`Harmony.GetOriginalMethod` -> `HarmonySharedState.GetOriginal`
`Harmony.GetMethodFromStackframe` -> `HarmonySharedState.FindReplacement` -> `HarmonySharedState.GetOriginal`

The problem was that `PlatformTriple.Current.GetIdentifiable` was only called from `Harmony.GetOriginalMethod`, so `Harmony.GetMethodFromStackframe` never used it and couldn't find the method. So I moved it from `Harmony.GetOriginalMethod` to `HarmonySharedState.GetOriginal`.

Also, and I only noticed this now, `HarmonySharedState.FindReplacement` actually returns the original and not the replacement, is this intentional? Because then `Harmony.GetMethodFromStackframe` actually returns the original, not the replacement as in docs, and `Harmony.GetOriginalMethodFromStackframe` will always fail, basically.
As far as I understand it, this is a bug, and then original test makes sense, not my version

pardeike:
I started reviewing your PR and created this branch. I think I cleaned up what is confusing. You can have a look at my changes - they explain it better than I could with words: <https://github.com/pardeike/Harmony/tree/fix/get-method-from-stackframe>

pardeike:
Sorry, I am going to reject this PR mainly because my own branch has a more clean design and fixes things that this does not address.

PR #622, opened 2024-08-30
Allows empty string for targeting special MethodType options and doesn't try searching for it with AccessTools

Banane9:
Allows patch attributes like `[HarmonyPatch(typeof(TargetType), "", MethodType.Getter)]` on a method, or
```cs
[HarmonyPatch(typeof(TargetType))]
static class PatchClass
{
    [HarmonyPatch("", MethodType.Getter)]
    static void Postfix() { }
}
```
to target an Indexer for example. With just `null` instead of `""`, the second example would be ambiguous.

This may allow making the targeting "nameless" members like (static) constructors or indexers more explicit, by allowing to provide an empty string in addition to the MethodType or `null`. Now that I'm writing this, it seems almost unnecessary, but the idea came while [talking about possible Analyzers for HarmonyPatch attributes](https://discord.com/channels/131466550938042369/674571535570305060/1278776582797918310)

For the AccessTools methods, it checks for empty string as well, as there can't be members with that name. Not with whitespace either, but net3.5 doesn't have that method.

### DIFF
--- a/Harmony/Internal/PatchTools.cs
+++ b/Harmony/Internal/PatchTools.cs
@@ -90,17 +90,17 @@ namespace HarmonyLib
 				switch (attr.methodType)
 				{
 					case MethodType.Normal:
-						if (attr.methodName is null)
+						if (string.IsNullOrEmpty(attr.methodName))
 							return null;
 						return AccessTools.DeclaredMethod(attr.declaringType, attr.methodName, attr.argumentTypes);
 
 					case MethodType.Getter:
-						if (attr.methodName is null)
+						if (string.IsNullOrEmpty(attr.methodName))
 							return AccessTools.DeclaredIndexerGetter(attr.declaringType, attr.argumentTypes);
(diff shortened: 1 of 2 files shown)


pardeike:
I get the idea and it seems reasonable to have it that way. Can you just clean up the diff at the end where there are empty line changes and also unnecessary “private”? I’ll pull it after that.

Banane9:
Ah, that was probably some auto-formatting, although I toggled it off 🤔

Won't be home again until Monday, so if you want to merge it immediately you can edit it too. At least that's toggled on, never used that feature myself before.

Banane9:
Okay, completely forgot about this, but here you finally go :D

pardeike:
Would you mind pushing an update that solves the conflict from your last PR?

Banane9:
> Would you mind pushing an update that solves the conflict from your last PR?

Done 👍

PR #633, opened 2024-11-15
MethodDelegate: allow for swapping value type with object, and allow ref for arg 0 on value type

Eddio0141:
This changes `AccessTools.MethodDelegate` to let you use `object` where a value type exists

Also lets you use `ref` for the first argument with value types, since currently you can't mutate structs from dynamic methods generated when open instance without boxing, so this lets you use delegates such as `void MyFunc(ref UnityEngine.Resolution res);`

### DIFF
--- a/Harmony/Tools/AccessTools.cs
+++ b/Harmony/Tools/AccessTools.cs
@@ -1524,6 +1524,13 @@ namespace HarmonyLib
 		/// else, invocation of the delegate calls the exact specified <paramref name="method"/> (this is useful for calling base class methods)
 		/// Note: if <c>false</c> and <paramref name="method"/> is an interface method, an ArgumentException is thrown.
 		/// </param>
+		/// <param name="delegateArgs">
+		/// Only applies for instance methods, and if argument <paramref name="instance"/> is null.
+		/// This argument only matters if the target <paramref name="method"/> signature contains a value type (such as struct or primitive types),
+		/// and your <typeparamref name="DelegateType"/> argument is replaced by a non-value type
+		/// (usually <c>object</c>) instead of using said value type.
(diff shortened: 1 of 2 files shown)


pardeike:
That’s nice work!

pardeike:
Maybe you can add some tests too?

Eddio0141:
Thanks! Also sorry for the messy commits, I've squashed them to two commits to try make it clean
I've added 2 tests which should test the exact code, it is all done now and should be passing

Also I just realised, this fix would only apply for `virtualCall: false`, since delegates is incompatible normally, and this fixes just the `dmd` to handle these things

pardeike:
Looks like we broke backwards compatibility with this PR. A user posted the following problem today on the discord:

> Hi, I'm having an issue with my mod that uses Harmony. My mod is compiled with Harmony 2.2.2 but the Unity Mod Manager that I'm using to inject my mod into the game has updated to using Harmony 2.3.5 which breaks compatibility.
> 
> If my mod compiled with Harmony 2.2.2 is run using Unity Mod Manager with Harmony 2.3.5 it can patch methods just fine but when the time comes to actually run the patched method I get the following error:
> ```
> MissingMethodException: Method not found: !!0 HarmonyLib.AccessTools.MethodDelegate<!0>(System.Reflection.MethodInfo,object,bool)
> ```
> 
> I can fix the error by simply compiling with Harmony 2.3.5 but that'll make the mod incompatible for anyone trying to use an older version of Unity Mod Manager with Harmony 2.2.2. I haven't changed the code. The only thing that's changed is the Harmony .dll file. So I was hoping anyone had a more robust solution?

Banane9:
Yup, adding a default parameter is source-, but not binary-compatible :C

Eddio0141:
Ah sorry about that, I didn't see this problem
I'm guessing #652 solves this issue

PR #637, opened 2024-12-17
update MonoMod.Core to 1.2.2 to address net9 issue

stiefeljackal:
This PR address the issue with Harmony when running under .NET 9. As @tyb-dev pointed out, .NET 9 changed LocalBuilder from `sealed partial class` to `abstract class`. Due to this, `MonoMod.Core` on `1.2.1` was failing because it was attempting to create an instance of LocalBuilder, resulting in the following:

```
System.MemberAccessException : Cannot create an instance of System.Reflection.Emit.LocalBuilder because it is an abstract class.
```

`MonoMod.Core` was updated to `1.2.2` roughly six days ago to address this update to `LocalBuilder`. Updating the package reference to `1.2.2` addresses this issue with Harmony, and all tests are now passing. There was one test that required an update due to type checking for `LocalBuilder` (now `RuntimeLocalBuilder` when on .NET 9).

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -3,7 +3,7 @@
 	<PropertyGroup>
 		<HarmonyVersion>2.3.3.0</HarmonyVersion>
 		<HarmonyPrerelease></HarmonyPrerelease>
-		<MonoModCoreVersion>1.2.1</MonoModCoreVersion>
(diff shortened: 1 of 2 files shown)


pardeike:
Thanks, last time I checked it was not released

PR #651, opened 2025-02-26
HarmonyPriority Documentation Update

garrettluskey:
Since I was confused whether a higher value meant higher priority. I updated it to inform others.


pardeike:
I don't get why this isn't obvious. Your additional wording seems obvious:

- Priority has a value
- that value can be high
- if the value of a Priority is high, then the Priority is high

How else would it be?

garrettluskey:
Agreed

PR #652, opened 2025-03-01
Fix issue that popped up with #633 and add string overload for new feature

Banane9:
Fixes the MissingMethod exception caused by #633 and adds the type-by-name overload for it as well.
Also a minor doc xml fix.


pardeike:
CI fails

Banane9:
Not sure why the CI is still failing - the error seems to be entirely unrelated to my changes and it works locally

pardeike:
I close this PR in favour of my commits that we discussed together in the Harmony discord

PR #765, opened 2026-06-22
Avoid importing unused assembly Serialization.Formatters in .NET 9.0+

aubymori:
The assembly 'System.Runtime.Serialization.Formatters.Binary' is banned by Paint.NET and attempting to load Harmony in Paint.NET through its effects system is impossible. Since this assembly isn't used in .NET 9.0+, I used the preprocessor to avoid loading it at all in .NET 9.0+. This allows Harmony to load in Paint.NET.

Here is what Paint.NET showed in its Plugin Errors section before this patch.
image

And after. The error comes from my DLL and not 0Harmony.dll being rejected.
image

### DIFF
--- a/Harmony/Serialization/PatchInfoSerialization.cs
+++ b/Harmony/Serialization/PatchInfoSerialization.cs
@@ -1,7 +1,9 @@
 using System;
 using System.IO;
 using System.Runtime.Serialization;
+#if !NET9_0_OR_GREATER
 using System.Runtime.Serialization.Formatters.Binary;
+#endif
 #if NET5_0_OR_GREATER
 using System.Text.Json;
 #endif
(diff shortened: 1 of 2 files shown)


pardeike:
Have you tried to use `Lib.Harmony.Thin`? It’s made for exactly this case where you want to control the dependencies yourself.

aubymori:
> Have you tried to use `Lib.Harmony.Thin`? It’s made for exactly this case where you want to control the dependencies yourself.

`System.Runtime.Serialization.Formatters` is not an external dependency but part of the .NET standard library. Regardless, `Lib.Harmony.Thin` still needlessly imports this assembly, causing Paint.NET to block it. Paint.NET will block *any* assembly with the name `System.Runtime.Serialization.Formatters`, so I think this patch is needed to allow Harmony to load in Paint.NET.

